PR #11, opened 2019-12-10
Single indicator & InfluenceMap weighting methodology 

vintented:
Still needs a few minor fixes and I have to finish preparing the sample datasets, but almost there!


maurolepore:
@jdhoffa, can you please comment on if and when we might work on this? It's been sitting here for a while and I'm unsure if I should close it for now.

jdhoffa:
I will close it. No idea when we will end up working on this (likely for a version 2) but in that case, it is also likely already stale and I will just work on a fresh re-write.

PR #54, opened 2020-05-18
50 join by scenario source

maurolepore:
I'm bringing this PR from my fork to here. I had pushed to my for while I waited for another PR to be merged first.


maurolepore:
I cleaneu up this PR. Follow up on #58

PR #56, opened 2020-05-18
Remove dependency on r2dii.scenario

maurolepore:


maurolepore:
Thanks! We are now only a short step away from passing CI.

PR #61, opened 2020-05-20
Use weo_2019

maurolepore:
Relates to 2degreesinvesting/r2dii.data#68


maurolepore:
This should fail until we merge 2DegreesInvesting/r2dii.data#68.

PR #72, opened 2020-05-28
join_ald_scenario() now warns if output has 0 rows

### LINKED ISSUE #65: Examples of all summarize_*() functions output zero rows
* https://2degreesinvesting.github.io/r2dii.analysis/reference/summarize_company_production.html#examples
* https://2degreesinvesting.github.io/r2dii.analysis/reference/summarize_portfolio_production.html#examples
* https://2degreesinvesting.github.io/r2dii.analysis/reference/summarize_weighted_production.html#examples

Here is a copy of one of those examples:

```r
library(r2dii.analysis)
library(r2dii.data)
library(r2dii.match)

master <- r2dii.data::loanbook_demo %>%
  r2dii.match::match_name(r2dii.data::ald_demo) %>%
  r2dii.match::prioritize() %>%
  join_ald_scenario(r2dii.data::ald_demo, r2dii.data::scenario_demo_2020)

summarize_company_production(master)#> # A tibble: 0 x 8
#> # … with 8 variables: sector <chr>, technology <chr>, year <dbl>,
#> #   name_ald <chr>, scenario <chr>, tmsr <dbl>, smsp <dbl>,
#> #   weighted_production <dbl>
summarize_company_production(master, use_credit_limit = TRUE)#> # A tibble: 0 x 8
#> # … with 8 variables: sector <chr>, technology <chr>, year <dbl>,
#> #   name_ald <chr>, scenario <chr>, tmsr <dbl>, smsp <dbl>,
#> #   weighted_production <dbl>
```

maurolepore:
Also closes #65


R/join_ald_scenario.R
```diff
@@ -42,14 +42,27 @@ join_ald_scenario <- function(data,
...
+
+  out
+}
+
+warn_if_has_zero_rows <- function(data) {
+  if (identical(nrow(data), 0L)) {
+    rlang::warn(class = "has_zero_row", "`data` has zero-rows.")
+  }
+
+  invisible(data)
```
maurolepore, line 65: I'm considering adding the new argument `message` to `warn_if_has_zero_rows()` so that we can throw different warnings at each step in the pipeline. 
Not yet sure -- might be an overkill.

R/join_ald_scenario.R
```diff
@@ -42,14 +42,27 @@ join_ald_scenario <- function(data,
   region_isos <- region_isos %>%
     dplyr::rename(scenario_source = .data$source)
 
-  data %>%
+  out <- data %>%
     left_join(ald, by = ald_columns()) %>%
```
maurolepore, line 46: @jdhoffa, do you confirm the type of join we want here is `left_join()`? 

I see we later use `inner_join()`. 

Compare:

``` r
library(dplyr, warn.conflicts = FALSE)

data <- tibble(name_ald = "a", sector = "x")
ald <-  tibble(name_ald = "b", sector = "y")

left_join(data, ald)
#> Joining, by = c("name_ald", "sector")
#> # A tibble: 1 x 2
#>   name_ald sector
#>   <chr>    <chr> 
#> 1 a        x

inner_join(data, ald)
#> Joining, by = c("name_ald", "sector")
#> # A tibble: 0 x 2
#> # … with 2 variables: name_ald <chr>, sector <chr>
```

<sup>Created on 2020-05-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jdhoffa: Yes, I confirm that this is correct. 
In the first case, we are linking the portfolio to any matches in the ald, hence we want to keep all matches in the loanbook (this is useful later for coverage calculations). 

In the second case, we are trying to filter the first dataframe based on the second, and we want only rows that are common to both

jdhoffa: Thanks for checking

tests/testthat/test-join_ald_scenario.R
```diff
@@ -221,8 +221,8 @@ test_that("include/excludes `plant_location`s inside/outside a region", {
   region_isos_toy <- tribble(
     ~region, ~isos, ~source,
     "north america", "cn", "demo_2020",
-    "oecd", "de", "demo_2020",
-    "oecd", "fr", "demo_2020",
+    "oecd_europe", "de", "demo_2020",
```
maurolepore, line 224: Good catch, but I wonder why you chose to add "_europe" here instead of removing it below. That would make this test consistent with other tests and with the dataset `r2dii.data::region_isos`. Any comment?

jdhoffa: no comment, your suggestion makes more sense!

Issue #67, opened 2020-05-28
Remove the `add_*()` prefix where output does not preserve original columns


maurolepore:
The `add_*()` functions here do not preserve the original columns

![image](https://user-images.githubusercontent.com/5856545/83146851-770f2b00-a0bc-11ea-80a4-e79241ab541a.png)

In dplyr, `add_count()` and `add_tally()` preserve the original columns and add a new one -- which  differentiates these functions from their summary counterparts `count()` and `tally()`. I like the idea of reserving the prefix "add_" to communicate that the output will have the same columns that the input has (plus something else).

Is this something we want here too?

``` r
library(dplyr, warn.conflicts = FALSE)

# In dplyr, summary functions don't preserve columns
# Original
mtcars %>% ncol()
#> [1] 11
# Summary
mtcars %>% count(cyl) %>% ncol()
#> [1] 2
mtcars %>% group_by(cyl) %>% summarize(n = n()) %>% ncol()
#> [1] 2

# But functions with the prefix "add_" preserve original columns
mtcars %>% ncol()
#> [1] 11
# Same columns plus one
mtcars %>% add_count(cyl) %>% ncol()
#> [1] 12
```

<sup>Created on 2020-05-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jdhoffa:
I like this idea. For now, let's just call these functions `portfolio_target` and `sda_target`, as I think the summary results are more useful/ easier understood by the audience at this stage. and we can add the add_* functions later, what do you think?

jdhoffa:
we also might consider renaming `add_portfolio_target` and `add_company_target` to `portfolio_production_target` and `company_production_target` to better distinguish from `sda_target`.

maurolepore:
How about excluding "production", in the same way `sda_target()` excludes "emission-factor"?

So the suite would be:

* `sda_target()`
* `portfolio_target()`
* `company_target()`

And if we go that route, then how about using `target_` as a prefix, so they are easier to find with tab-complete. So:

* `target_sda()`
* `target_portfolio()`
* `target_company()`

jdhoffa:
I propose calling the functions: 
* `target_sda()`
* `target_fair_share_portfolio()`
* `target_fair_share_company()`

[...]

PR #77, opened 2020-05-29
Remove `add_` prefix (#67)

jdhoffa:
I plan to address a couple issues here, but for now I have:
* Renamed all functions to remove the `add_` prefix, and given them more evocative names
* Renamed the vignettes to more easily identify the corresponding function.


jdhoffa:
Latest commit:
Closes #67

R/target_market_share_company.R
```diff
@@ -11,6 +11,8 @@
 #'   `tmsr_target_weighted_production` and `smsp_target_weighted_production`.
 #' @export
 #'
+#' @family calculate scenario targets
```
maurolepore, line 14: I'll add "functions to" so this reads better. 

Compare:

image

versus

image

jdhoffa: :)

PR #85, opened 2020-05-29
Update readme

maurolepore:
Render readme to reflect new outputs in long format.


maurolepore:
This is a small change to syncronize README. Maybe I should try to setup the [GitHub Action to render README.md](https://github.com/r-lib/actions/blob/master/examples/render-readme.yaml). I tried it once and failed. But the file has been updated recently, and maybe the new versiono "just works".

jdhoffa:
Do you want me to leave this open for now so you can see if we can automate this in an action?

maurolepore:
Yeah, I'll try today.

maurolepore:
I'll update redme in another pr.

PR #89, opened 2020-06-03
Pick tmsr or smsp target based on technology (#82)

### LINKED ISSUE #82: Pick smsp or tmsr target based on technology
* Add this dataset as internal data: 

sector | technology | which_metric
-- | -- | --
automotive | electric | smsp
automotive | hybrid | smsp
automotive | ice | tmsr
automotive | fuelcell | smsp
power | hydrocap | smsp
power | renewablescap | smsp
power | coalcap | tmsr
power | gascap | tmsr
power | oilcap | tmsr
power | nuclearcap | smsp
oil and gas | oil | tmsr
oil and gas | gas | tmsr
coal | coal | tmsr

* in the `target_market_share_*` functions, at the end of the function, pick the appropriate target values based on the specification in this file.

### LINKED ISSUE #86: Brainstorm new (name,value) pairs for the output of `target_*`
Relates to https://github.com/2DegreesInvesting/r2dii.analysis/pull/81
> Looks good. I agree we need to polish names -- not only the names of the columns but also the name of the values that get into the pivoted columns. But that's another discussion. Maybe we could do a brainstorm session and involve other folks.

jdhoffa:
* Add new internal dataset `pick_tmsr_or_smsp` which bridges the sector+technology to the appropriate scenario metric (tmsr for high-carbon, smsp for low-carbon)
* Output of `target_*` is now long-format
* Choose more evocative name for columns and values (#86) might need more workshopping 

Closes #82 
Maybe closes #86


R/target_market_share_company.R
```diff
@@ -97,9 +97,27 @@ target_market_share_company <- function(data) {
       )
     ) %>%
     tidyr::pivot_longer(
-      cols = tidyr::ends_with("weighted_production"),
-      names_to = "production_name",
-      values_to = "production_value"
+      cols = c("tmsr_target_weighted_production", "smsp_target_weighted_production"),
```
maurolepore, line 100: Just noticing that these two files seem to have common patterns that we might be able to DRY:

* R/target_market_share_company.R
* R/target_market_share_portfolio.R

jdhoffa: Ya absolutely, let's open this as a new issue and discuss, prior to submitting to CRAN. Overall I think there are a couple of places we could do some housekeeping in this package. 
In particular, `summarize_portfolio_production` and `summarize_company_production` both exist, because I think both outputs are useful to the banks. But because we have two pathways, we have two target functions etc. etc. 

Likely we could just add a `...` argument to the function, and allow for a "company" grouping variable, to achieve both results, and then add a vignette to explain how to do this.

maurolepore:
Okay, seems to work fine. I'll go ahead and merge and let this PR close #82.
I think we should keep https://github.com/2DegreesInvesting/r2dii.analysis/issues/86 open. I'll make some comments there.

Issue #93, opened 2020-06-05
Include wrapper to calculate a "market share benchmark" based on the `ald`


jdhoffa:
Currently the output of `target_market_share_*` includes `c("portfolio", "target_*",...)`. 
We need to include a benchmark value, which is just the normalized regional sum of production of the `ald`. 

This should be relatively easy to extract from `summarize_weighted_production`. 
@maurolepore Maybe we should do this after the interface is a little more stable?

PR #97, opened 2020-06-15
add ald benchmark (#93)

jdhoffa:
Added a module in `target_market_share` to calculate the, region specific, sum of ALD production, normalized to the starting portfolio value. 

- [x] Still need to add some more tests to ensure output is correct


jdhoffa:
Closes #93

PR #104, opened 2020-06-17
ald benchmark ult owner (#103)

### LINKED ISSUE #103: In `target_market_share` when calculating normalized_ald_benchmark, filter for `is_ultimate_owner == TRUE`
This ald-based benchmark is meant to reflect the actual production of all companies in the corporate universe (in the ald). Since we have companies at various levels in the ALD, we must aggregate data only at ultimate owner, otherwise we will be double counting production. 

Also rename `normalized_ald_benchmark` to `corporate_economy`. 

Thanks @2diiKlaus

jdhoffa:
@maurolepore I have made a couple pretty small changes regarding a mistake I had made that @2diiKlaus pointed out. 
I also changed the name of the benchmark output of `target_market_share` from `normalized_ald_benchmark` to `normalized_corporate_economy` as per Klaus' comments. 

On the note of nomenclature, I think that these metric names are evocative and descriptive, however I think it would make sense to explain what these values are in a little more detail, where do you think the best place to host this is? I think we had briefly discussed the idea of a data_dictionary for the output, maybe this is the time to start thinking about that? 

Closes #103


maurolepore:
In the gh-actions workflow I see a  step that fails but I would merge anyway. The causes of the failure are likely external; and the problem appears with R 3.5 only. 

image

maurolepore:
Closes #103

* Changes following a discussion with @2diiKlaus

Also:

`normalized_ald_benchmark ---> normalized_corporate_economy`

Issue #113, opened 2020-06-24
Release r2dii.analysis 0.0.1


maurolepore:
Prepare for release: 
 
* [x] Check that description is informative 
* [x] Check licensing of included files 
* [x] `devtools::build_readme()` 
* [x] `usethis::use_cran_comments()` 
* [x] `devtools::check(remote = TRUE, manual = TRUE)` 
* [x] `devtools::check_win_devel()` 
  https://win-builder.r-project.org/dkHrN77skGd1/00check.log
* [x] `rhub::check_for_cran()` 
   https://builder.r-hub.io/status/r2dii.analysis_0.0.0.9004.tar.gz-051f670e458142ae805b32f741c803bc
   On windows I see this error but I think it's unimportant; I think it only appears on rhub, and not on win-builder -- which is what really matters.
  ```r
   Error in loadNamespace(name) : there is no package called 'utf8
   ```
   https://builder.r-hub.io/status/r2dii.analysis_0.0.0.9004.tar.gz-7d3efe03030642b38f130b3fc1e8a9b1
   https://builder.r-hub.io/status/r2dii.analysis_0.0.0.9004.tar.gz-c0f34e3c03474c48a063b681fbebf828
* [x] Update `cran-comments.md` 
* [x] Review pkgdown reference index for, e.g., missing topics 
 
Submit to CRAN: 
 
* [x] `usethis::use_version('patch')` 
* [x] `devtools::submit_cran()` 
* [x] Approve email 
 
Wait for CRAN... 
 
* [x] Accepted :tada: 
* [x] `usethis::use_news_md()` 
* [x] `usethis::use_github_release()` 
* [x] `usethis::use_dev_version()` 
* [x] Update install instructions in README

jdhoffa:
r2dii.analysis is on CRAN!

I have completed this checklist, closing this issue now.

Issue #107, opened 2020-06-18
Add demo article, including visualizations, to pkgdown articles


jdhoffa:
I use this issue to initiate a discussion on visualization as I draft an .Rmd to show plots using the latest version of `r2dii.analysis`. This process helps to identify potential bugs and annoying user interfaces along the way.

I will aim to extract the .Rmd as an article to host on the analysis webpage. 

No need to address anything here, I will close the issue when I am happy.

jdhoffa:
FYI @maurolepore here is my solution to programatically draw in ribbons. It's not the most elegant, but I think it actually works pretty well, have a look and see if it can be distilled into something similar (noting that in general the ordering of target_1, 2, 3 etc could be completely arbitrary):

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(ggplot2)

data <- tibble::tribble(
               ~name, ~year, ~value,
          "target_1", 2020L,     0L,
          "target_1", 2021L,     1L,
          "target_1", 2022L,     2L,
          "target_1", 2023L,     3L,
          "target_1", 2024L,     4L,
          "target_1", 2025L,     5L,
          "target_2", 2020L,     0L,
          "target_2", 2021L,     2L,
          "target_2", 2022L,     4L,
          "target_2", 2023L,     6L,
          "target_2", 2024L,     8L,
          "target_2", 2025L,    10L,
          "target_3", 2020L,     0L,
          "target_3", 2021L,     4L,
          "target_3", 2022L,     8L,
          "target_3", 2023L,    12L,
          "target_3", 2024L,    16L,
          "target_3", 2025L,    20L,
          "target_4", 2020L,     0L,
          "target_4", 2021L,     8L,
          "target_4", 2022L,    16L,
          "target_4", 2023L,    24L,
          "target_4", 2024L,    32L,
          "target_4", 2025L,    40L
          )

ceiling <- data %>% 
  group_by(year) %>% 
  summarize(name = "target_ceiling",
         value = max(value)) %>% 
  ungroup() %>% 
  mutate(value = max(value))
#> `summarise()` ungrouping output (override with `.groups` argument)
  

data %>% 
  rbind(ceiling) %>% 
  group_by(year) %>% 
  arrange(year, value) %>% 
  mutate(
[...]

PR #119, opened 2020-06-25
in target_* function, drop groups and warn if input data is grouped

### LINKED ISSUE #114: target_sda() outputs a grouped data frame. Is this intentional?
``` r
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

valid_matches <- match_name(loanbook_demo, ald_demo) %>%
  prioritize()

out <- valid_matches %>%
  target_sda(
    ald = ald_demo,
    co2_intensity_scenario = co2_intensity_scenario_demo
  )

dplyr::is_grouped_df(out)
#> [1] TRUE
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### LINKED ISSUE #117: target_market_share() may need to drop groups
In general I expect a function to preserve groups, but that is impossible if the output excludes the grouping variable.  This example throws an error because the grouping variable in the input `data` is not present in the output.

``` r
library(r2dii.match)
library(r2dii.analysis)
devtools::load_all()
#> Loading r2dii.analysis

grouped_by_column_not_in_output <- fake_matched() %>% 
  dplyr::group_by(.data$id_loan)

target_market_share(
  grouped_by_column_not_in_output,
  fake_ald(),
  fake_scenario(),
  r2dii.data::region_isos_demo
)
#> Error: Must group by variables found in `.data`.
#> * Column `id_loan` is not found.
```

<sup>Created on 2020-06-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

For all functions that return columns different than those passed to data, it is hard to promise we'll keep groups. I suggest we drop them with a warning. This will keep our promise simple, and the code simpler than it currently is.

jdhoffa:
Closes #114 
Closes #117


jdhoffa:
I have elected the more strict option: 
- ungroup input data
- warn when doing so
- output ungrouped data. 

I can see a use-case where we may want to revisit this choice. As a concrete example: a financial supervisor may want to run multiple loanbooks. It would be easy to just add a column title "loanbook_name" and group by this to calculate targets per loanbook. That said, it is a little tricky because there are obviously internal groupings happening etc. so we would have to be clever about which input groupings are allowed (we could warn if any crucial columns are grouped, and ungroup only those columns or something like this). I think this is something we should aim for in version-0.0.2. 

For now, I'd like to prioritize #107 and #113

R/target_market_share.R
```diff
@@ -4,6 +4,8 @@
 #' calculated using the market share approach applied to each relevant climate
 #' production forecast.
 #'
+#' @template ignores-existing-groups
```
maurolepore, line 7: Well done!

R/target_market_share.R
```diff
@@ -62,9 +64,7 @@ target_market_share <- function(data,
                                 by_company = FALSE) {
   stopifnot(is.data.frame(data))
 
-  old_groups <- dplyr::groups(data)
-  data <- ungroup(data)
-
+  data <- ungroup_and_warn(data, "Ungrouping input data.")
```
maurolepore, line 67: :+1:

R/target_market_share.R
```diff
@@ -174,7 +174,7 @@ target_market_share <- function(data,
       names_to = "weighted_production_metric",
       values_to = "weighted_production_value"
     ) %>%
-    group_by(!!!old_groups)
+    ungroup()
```
maurolepore, line 177: :+1:

tests/testthat/test-target_market_share.R
```diff
@@ -16,6 +16,30 @@ test_that("outputs a tibble", {
   expect_is(out, "tbl_df")
 })
 
+test_that("outputs is ungrouped", {
```
maurolepore, line 19: :1st_place_medal:

maurolepore:
Looks great. Please review my changes and merge when you are ready.

PR #152, opened 2020-07-29
Allow tidyverse

maurolepore:
I expect this PR to fix this problem:

image


maurolepore:
This should be low risk so I'll go ahead and merge.

Issue #118, opened 2020-06-25
Add `by_company` flag to `target_sda` function to allow for company level results


jdhoffa:
Exactly analogous to how `target_market_share(..., by_company = T)` outputs market share targets at company level, we should add a flag to do this in `target_sda()`.

Thanks @vintented 

```r
library(r2dii.analysis)
library(r2dii.data)
library(r2dii.match)

match_result <- r2dii.data::loanbook_demo %>%
  r2dii.match::match_name(r2dii.data::ald_demo) %>%
  r2dii.match::prioritize()

# we already have this
# calculate targets at company level
target_market_share(match_result,
                    ald = r2dii.data::ald_demo,
                    scenario = r2dii.data::scenario_demo_2020,
                    region_isos = r2dii.data::region_isos_demo,
                    by_company = TRUE
)

# want to have something like this with similar outputs
# calculate sda targets at company level
target_sda(match_result,
                    ald = r2dii.data::ald_demo,
                    scenario = r2dii.data::co2_intensity_scenario_demo,
                    by_company = TRUE
)
```

PR #155, opened 2020-08-03
`target_sda()` gains `by_company` argument

jdhoffa:
`by_company` is a logical argument. 
When `TRUE`, it outputs SDA targets calculated at the company level.


jdhoffa:
Closes #118

PR #164, opened 2020-08-05
Check that `id_loan` is not duplicated

### LINKED ISSUE #160: `projected` and `target_*` values for `steel` output of `target_sda()` seems off
From a conversation with a beta-tester:
> Something seems to be off with the steel projected and target output (very high)

sector | year | emission_factor_metric | emission_factor_value
-- | -- | -- | --
steel | 2019 | projected | 4.872001912
steel | 2019 | corporate_economy | 1.867306771
steel | 2019 | target_b2ds | 4.872001912
steel | 2019 | adjusted_scenario_b2ds | 1.867306771

> I checked the CO2 intensity values of matched companies and they are all less than 4:

year | min_value | max_value
-- | -- | -- 
2019 | 0.0236 | 3.02
2020 | 0.0236 | 3.02
2021 | 0.0236 | 3.02
2022 | 0.0236 | 3.02
2023 | 0.0236 | 3.02
2024 | 0.0236 | 3.02
2025 | 0.0236 | 3.02

---

I am unable to reproduce this error as of now, but will try again and post here as soon as I have.

jdhoffa:
This relates to a bug identified by multiple users, where `projected` `emission_factors` were unreasonably large. 

This was a result of the input data having duplicate `id_loans` per loan. The result was that, while the loan weights was calculated per `id_loan`, the `emission_factor` was summed across each loan weight, causing double-counting (sometimes more). 

I decided to go the more harsh route, and cause the function to `error` if input `id_loans` were duplicated. This will save us headache down the road, as these `id`s should uniquely identify EVERY loan (otherwise we're gonna have problems). 

I also noticed that the `emission_factor` was still being weighted when `by_company = TRUE` which doesn't make sense, so fixed that as well. 

Maybe closes #160

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -16,6 +16,8 @@ check_unique_id <- function(data, column){
       sprintf("Column `%s` must not contain any duplicates.", column)
     )
   }
+
+  invisible(data)
 }
 
 warn_grouped <- function(data, message) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -9,7 +9,7 @@ check_no_value_is_missing <- function(data, column) {
   invisible(data)
 }
 
-check_unique_id <- function(data, column){
+check_unique_id <- function(data, column) {
   if (sum(duplicated(data[[column]]))) {
     abort(
       class = "unique_ids",
--- a/R/utils.R
+++ b/R/utils.R
@@ -9,6 +9,15 @@ check_no_value_is_missing <- function(data, column) {
   invisible(data)
 }
 
+check_unique_id <- function(data, column){
+  if (sum(duplicated(data[[column]]))) {
+    abort(
+      class = "unique_ids",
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -9,6 +9,15 @@ check_no_value_is_missing <- function(data, column) {
...
 }
 
+check_unique_id <- function(data, column){
+  if (sum(duplicated(data[[column]]))) {
+    abort(
+      class = "unique_ids",
+      sprintf("Column `%s` must not contain any duplicates.", column)
+    )
+  }
+}
```
maurolepore, line 19: This function is called for its side effect; when the call to `if()` is FALSE we need to return `invisible(data)` (https://principles.tidyverse.org/out-invisible.html).

maurolepore: Done it.

PR #175, opened 2020-08-13
Now pulls dataset `green_or_brown` from `r2dii.data`

### LINKED ISSUE #146: Transfer green_or_brown to .data
Once we merge 2DegreesInvesting/r2dii.data#124, we should: 
- [x] Reference `green_or_brown` from .data instead of internal data in .analysis
- [x] Remove `green_or_brown` from "internal data" of .analysis
- [x] Move the tmsr_or_smsp mapping from internal data, into `R/utils.R` (the file is very small, I think it's fine to write the mapping as a vector instead of an internal dataset)

jdhoffa:
Biggest change here is that we now have to import `r2dii.data`, instead of suggesting it. 

Closes #146


jdhoffa:
Good to merge?

PR #181, opened 2020-08-19
Add option to output unweighted production

### LINKED ISSUE #131: Option to not output "portfolio-weighted" market_share targets
As a bank, I would like to be able to calculate the output of `target_market_share(..., by_company = FALSE)`, WITHOUT multiplying `production` by `loan_size_*` at all. 

(We would refer to this as something like "client-level view", or "portfolio-level output, without portfolio-weighting")

### LINKED ISSUE #165: `target_market_share` output should not be `weighted_production_value` if `by_company = F`
This would likely just cause confusion, as we should not be loan_weighting at company level in the first place (note to self: make sure in the code that we are not doing this).

jdhoffa:
Closes #131 and closes #165

A couple of things: 
* This could probably use more tests. 
* The warning message I wrote is pretty verbose. Happy to rework that quite a lot, as I wasn't sure the best way to convey the message concisely. 
* The warning could potentially be an article, explaining why you would or wouldn't want to do these things.


jdhoffa:
For context, from our call recently:
Screen Shot 2020-08-19 at 3 09 04 PM


Note that @2diiKlaus advised not to worry about the parameter for `target_sda`, so this should affect only `target_market_share`.

R/target_market_share.R
```diff
@@ -53,13 +56,38 @@
 #'     region_isos = region_isos_demo,
 #'     by_company = TRUE
 #'   )
+#'
+#' # Calculate unweighted targets
```
maurolepore, line 60: Maybe move right on top of `weight_production`

R/target_market_share.R
```diff
@@ -53,13 +56,38 @@
...
+    is.data.frame(ald),
+    is.data.frame(scenario),
+    is.data.frame(region_isos),
+    is.logical(use_credit_limit),
+    is.logical(by_company),
+    is.logical(weight_production)
+  )
+
+  if (by_company & weight_production) {
+    warn(message = "You have selected to view the weighted production at company
```
maurolepore, line 86: Think if it makes sense to create a new class of warning.

R/target_market_share.R
```diff
@@ -53,13 +56,38 @@
...
+         level. Make sure this is what you want to see! It is often desirable to
+         see these results unweighted, to understand where the company actualyl
+         stands in the market. ")
```
maurolepore, line 89: * Try rewrite with tidyverse style: problem, hint.
* Maybe it's time to import glue.

R/target_market_share.R
```diff
@@ -80,12 +108,28 @@ target_market_share <- function(data,
...
-    ) %>%
-    add_ald_benchmark(ald, region_isos, by_company)
+    )
+
+  if (weight_production) {
+    data <- data %>%
+      summarize_weighted_production(
+        !!!rlang::syms(summary_groups),
+        use_credit_limit = use_credit_limit
+        ) %>%
```
maurolepore, line 118: May need to /style

R/target_market_share.R
```diff
@@ -80,12 +108,28 @@ target_market_share <- function(data,
...
+  } else {
+    data <- data %>%
+      group_by(
+        .data$sector_ald,
+        .data$technology,
+        .data$year,
+        !!!rlang::syms(summary_groups)
+        ) %>%
+      summarize(
+        weighted_production = sum(.data$production) # this is not actually "weighted"
```
maurolepore, line 129: Great use of comments!

maurolepore, line 129: Maybe we can capture the comment on a helper, then extract the common pattern in both branches of the if else.

R/target_market_share.R
maurolepore: Do we need a warning when both are FALSE?
image

jdhoffa: Ah woops, no need for a warning when both are FALSE! You may have heard Klaus refer to this as the "client-view". It's not the default, but it's useful

R/target_market_share.R
```diff
@@ -83,10 +83,14 @@ target_market_share <- function(data,
...
-    warn(message = "You have selected to view the weighted production at company
-         level. Make sure this is what you want to see! It is often desirable to
-         see these results unweighted, to understand where the company actualyl
-         stands in the market. ")
+    warn(
+      class = "incompatible_by_company_and_weight_production",
+      glue(
+        "`by_company` & `weight_production` shouldn't be both TRUE or FALSE`:
+        You've supplied `by_company = TRUE` and `weight_production = TRUE`.
+        Did you forget to set one of these arguments to `FALSE`?"
```
jdhoffa: This is a bit too strict of a warning for me. They shouldn't both be TRUE unless you are sure you want them to be, and that has to do with the context of what you want to analyze...

R/target_market_share.R
```diff
@@ -108,44 +108,34 @@ target_market_share <- function(data,
...
+      } else {
+        summarize_unweighted_production(
+          .,
+          .data$sector_ald,
+          .data$technology,
+          .data$year,
+          !!!rlang::syms(summary_groups)
+        )
+      }
+    } %>%
```
jdhoffa, line 127: I DIDN'T KNOW YOU COULD DO THIS!

R/target_market_share.R
```diff
@@ -85,9 +85,9 @@ target_market_share <- function(data,
   if (by_company & weight_production) {
     warn(
       glue(
-        "`by_company` & `weight_production` shouldn't be both TRUE`:
-        You've supplied `by_company = TRUE` and `weight_production = TRUE`.
-        Did you forget to set one of these arguments to `FALSE`?"
+        "You've supplied `by_company = TRUE` and `weight_production = TRUE`.
+        This will result in company-level, weighted by the portfolio loan size.
+        Did you mean to set one of these arguments to `FALSE`?"
```
jdhoffa, line 90: @maurolepore let me know how you feel about this instead. Just want to give a bit of context as to "why" this might be cause for concern.

jdhoffa: I'm going to wait until you're up and have a look at this, but otherwise happy to merge this.

R/target_market_share.R
maurolepore: Me neither but it works :) 
If it's too weird, wipe it out.

R/target_market_share.R
maurolepore: I see little indication that this is a concern. As it's too strong to say that this _should not_ happen, how about saying that "this is rarely useful".

jdhoffa: That sounds good to me.

PR #193, opened 2020-09-02
aggregate by `country_of_domicile`

### LINKED ISSUE #171: `target_sda()` results STILL don't make sense
After a call and some debugging with @QianFeng2020, we determined that there is still a bug in the `loanbook_with_weighted_emission_factors()` portion of the code, incorrectly weighting production values for the cement sector, and producing some VERY large emission factors. 

I don't have a lot of context other than this, but will explore myself with some fake data and see what I can reproduce. \

Relates to #160 STILL

jdhoffa:
So after a bit of digging, I think the error was actually just coming from an issue with how we aggregate the ALD from technology level to sector level. 
If additional columns are present, and their values differ, they cause double counting in the aggregation of the emission factors. This was the case with the `country_of_domicile` column. 

My solution was rather strict, and works only with that column because in principle, it's hard to say for sure how to aggregate the ALD in relation to differing columns. 

The reprex from #171 now looks good: 
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(ggplot2)
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

match_result <- loanbook_demo %>% 
  match_name(ald_demo) %>% 
  prioritize()

match_result$name_ald[1] <- "jsw group"
match_result$sector_ald[1] <- "steel"

out <- match_result %>% 
  target_sda(ald_demo,
             co2_intensity_scenario_demo) %>% 
  filter(sector == "steel", year >= 2020)

ggplot(
  data = out,
  mapping = aes(
    x = year,
    y = emission_factor_value,
    color = emission_factor_metric)
) +
  geom_line() +
  facet_wrap(~ sector)
```

<sup>Created on 2020-09-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Closes #171


maurolepore:
Relates to #171 

LGTM as it fixes a bug in a seemingly common case -- when `country_of_domicile` has multiple values.

Nit: We might need to extend this solution to handle any column.

PR #198, opened 2020-09-03
prepare 0.1.0

jdhoffa:
* update cran comments with revdepcheck results
* add revdep/ directory


maurolepore (approved):
LGTM. 

Nit: I'm not sure it it's best to include or exclude revdep/ from the git repo. It adds a lot of noise that may obscure the changes that really impact the workings of the packages. But I have not explored this issue deep enough to know what's best practice.

maurolepore:
Okay, I see the tidyverse team check in the revdep/ directory. And they seem to use it to keep track of the problems they see and need to be addressed in other packages.

Long story short, best practice is to commit revdep/.

PR #204, opened 2020-09-09
Outputs correctly with different scenarios per region

### LINKED ISSUE #203: `target_market_share()` should handle input scenario with different `scenario` value in each `region`
Consider the case where we have two scenarios (e.g. "sds" and "cps") defined on the "global" region, and only one scenario (e.g. "sds") defined for the "europe" region. The code should be robust to this, and only calculate targets if a scenario is available. 

Currently, it outputs a mixture of "list" outputs, and `NA`s. See the reprex below:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

matched <- tibble::tribble(
                  )

ald <- tibble::tribble(
                            "shaanxi auto", "automotive",       "ice",  2025,           1,                1,            "FR",               TRUE
                            )

scenario <- tibble::tribble(
                                 "sds", "automotive",       "ice", "europe",  2025,   0.5, -0.08,      "weo_2019",
                                 "sds", "automotive",       "ice", "global",  2025,   0.5, -0.08,      "weo_2019",
                                 "cps", "automotive",       "ice", "global",  2025,   0.5, -0.08,      "weo_2019"
                             )

out <- target_market_share(
  matched,
  ald,
  scenario
)
#> Warning: Values are not uniquely identified; output will contain list-cols.
#> * Use `values_fn = list` to suppress this warning.
#> * Use `values_fn = length` to identify where the duplicates arise
#> * Use `values_fn = {summary_fun}` to summarise duplicates
```

<sup>Created on 2020-09-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Thanks @jacobvjk

jdhoffa:
Closes #203

### DIFF
--- a/tests/testthat/test-target_market_share.R
+++ b/tests/testthat/test-target_market_share.R
@@ -446,3 +446,35 @@ test_that("with known input outputs `technology_share` as expected (#184)", {
     tolerance = 1e-3
   )
 })
+
+test_that("outputs sensibly with different scenarios per region (#203)", {
+  matched <- fake_matched()
+
+  ald <- fake_ald()
+
+  scenario <- fake_scenario(
+    scenario = c("cps", "sds", "sds"),
+    region = c("global", "global", "non opec"),
+    scenario_source = "weo_2019"
+  )
+
+  out <- target_market_share(
+    fake_matched(),
+    fake_ald(),
+    scenario
+  )
+
+  target_cps <- out %>%
+    filter(metric == "target_cps") %>%
+    split(.$region)
+
+  target_sds <- out %>%
+    filter(metric == "target_sds") %>%
+    split(.$region)
+
+  expect_equal(target_cps$global, 0.5)
+  expect_equal(target_sds$global, 0.5)
+  expect_equal(target_sds$`non opec`, 0.5)
+
+
+})


R/target_market_share.R
```diff
@@ -235,8 +235,11 @@ target_market_share <- function(data,
 
   data %>%
     pivot_longer(cols = starts_with("weighted_")) %>%
+    filter(!is.na(.data$value)) %>%
```
maurolepore, line 238: @jdhoffa, can you confirm this is reasonable? It seems to me but I'd like to double check as it may result in the loss of data.

R/target_market_share.R
```diff
@@ -235,8 +235,11 @@ target_market_share <- function(data,
...
     separate_metric_from_name() %>%
-    pivot_wider(names_from = .data$name) %>%
+    pivot_wider(names_from = .data$name, values_fn = list) %>%
+    tidyr::unnest(.data$production) %>%
+    tidyr::unnest(.data$technology_share) %>%
```
maurolepore, line 242: FYI, this is the same kind of solution we used above.

tests/testthat/test-target_market_share.R
```diff
@@ -446,3 +446,21 @@ test_that("with known input outputs `technology_share` as expected (#184)", {
...
+  )
+
+  out <- target_market_share(fake_matched(), fake_ald(), scenario)
+
+  target_cps <- out %>% filter(metric == "target_cps") %>% split(.$region)
+  target_sds <- out %>% filter(metric == "target_sds") %>% split(.$region)
+
+  expect_equal(target_cps$global$production, 0.5)
+  expect_equal(target_sds$global$production, 0.5)
+  expect_equal(target_sds$`non opec`$production, 0.5)
```
maurolepore, line 465: This tests are a little different from the ones you proposed, but I'm pretty sure what you meant is this.

maurolepore (commented):
@jdhoffa, whenever you can, see if this does the trick.

jdhoffa:
Yup, this looks perfect @maurolepore  thanks. 
Afaik no good information is lost with that `filter(!is.na())`.
Unfortuantely this is my PR, so I can't approve it, but it all lgtm

PR #230, opened 2020-11-26
Add internal `loanbook_stable` and `region_isos_stable` (#227)

### LINKED ISSUE #227: Are these differences expected?
The intended release of r2dii.data 0.1.5 exposed problems with reverse dependencies: r2dii.match and r2dii.analysis. 
The problems in r2dii.match seems related to changes in the column `borderline` 

``` r
subset(
  r2dii.data::data_dictionary, 
  subset = column == "borderline", 
  select = c("dataset", "column")
)
#>                   dataset     column
#> 14     cnb_classification borderline
#> 30    gics_classification borderline
#> 38    isic_classification borderline
#> 63    nace_classification borderline
#> 68   naics_classification borderline
#> 91 sector_classifications borderline
#> 95     sic_classification borderline
```

<sup>Created on 2020-11-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Here I see many differences and I need help to decide if updating the regression references is the right thing to do.

image

image

image

image

jdhoffa:
Closes #227

Relates to https://github.com/2DegreesInvesting/r2dii.match/pull/328

With these internal datasets, we will have a more stable output from regression tests, that depend less on unrelated changes to `r2dii.data`.


maurolepore (commented):
I'm thinking that maybe it is too obscure to overwrite the name `loanbook_demo` so deep into the code base. An alternative would be to use another name, e.g. `loanbook_stable` and use it one of these ways:

* To replace the calls to `loanbook_demo` (most transparent; many lines of code change).
```r
# r2dii.analysis/tests/testthat/test-*.R
match_name(loanbook_stable, ald_demo)
```

* To overwrite `loanbook_demo` but in a more obvious place, like at the top of a test-*.R file (a bit more transparent; only one line of code changes).
```r
# r2dii.analysis/tests/testthat/test-*.R
loanbook_demo <- loanbook_stable
```

* To output  `loanbook_stable` indirectly, via an interface (many lines of code change but the code becomes more maintainable because if we later want to output something different we only need to change the implementation of the interface).

```r
# r2dii.analysis/tests/testthat/helper-fake_lbk.R
fake_lbk <- function() {
  loanbook_stable
}

# r2dii.analysis/tests/testthat/test-*.R
match_name(fake_lbk(), ald_demo)
```

jdhoffa:
> One thing to consider is if we can find an intermediate refactoring step that keeps the tests passing at all times. I see CI is failing now.

They are expected to fail until 2DegreesInvesting/r2dii.match#327 is merged. Not sure if it's worth our time for an intermediate refactoring? what do you think?

jdhoffa:
Or I guess, are you worried about reverse dependencies on `r2dii.match`?

maurolepore:
Yeah, I'm likely overthinking -- in part because I'm looking for real-world opportunities to apply what I'm learning in "Working effectively with legacy code" as I read it. Sorry you are a victim of this process :)

maurolepore (approved):
@jdhoffa, have a look and edit or merge when you are ready. This should come after
https://github.com/2DegreesInvesting/r2dii.match/pull/328.

jdhoffa:
@maurolepore so this LGTM, but I have a question. 
This PR will cause the dev version of `r2dii.analysis` to fail with the current CRAN release of `r2dii.match`. Does this mean `r2dii.analysis` now depends on `r2dii.match (>= 0.0.7)`? Should we bypass these tests if `r2dii.match (<= 0.0.6)` for stability?

maurolepore:
> Should we bypass these tests if r2dii.match (<= 0.0.6) for stability?
I like this idea. I'm going to explore it.

maurolepore:
In order to decide if we need to skip the test, I'm unsure if the best condition is that r2dii.match <= 0.0.6. The problem is that some commit where r2dii.match = 0.0.6.9000 may lack the ability to allow reserved columns. Instead, a more direct way may be to check that `allow_reserved_columns()` exists in r2dii.match -- whatever the version. This is a bit fragile because `allow_reserved_columns` is internal, but I would trust our tests.


``` r
packageVersion("r2dii.match")
#> [1] '0.0.6.9000'

# https://bit.ly/2VksIPj (@hadly at StackOverflow)
allows_reserved_columns <- exists(
  'allow_reserved_columns',
  where = asNamespace('r2dii.match'),
  mode = 'function'
)

# We can use skip_if_not()
allows_reserved_columns
#> [1] TRUE

```

``` r
packageVersion("r2dii.match")
#> [1] '0.0.6'

# https://bit.ly/2VksIPj (@hadly at StackOverflow)
allows_reserved_columns <- exists(
  'allow_reserved_columns',
  where = asNamespace('r2dii.match'),
  mode = 'function'
)

# We can use skip_if_not()
allows_reserved_columns
#> [1] FALSE
```

<sup>Created on 2020-11-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

tests/testthat/test-summarize_weighted_production.R
```diff
@@ -186,7 +186,17 @@ test_that("preserves groups passed to ...", {
 })
 
 test_that("with demo data returns known value", {
-  master <- prioritize(match_name(loanbook_demo, ald_demo)) %>%
+  allows_reserved_columns <- exists(
+    'allow_reserved_columns',
+    where = asNamespace('r2dii.match'),
+    mode = 'function'
+  )
+  skip_if_not(allows_reserved_columns)
```
maurolepore, line 194: @jdhoffa this is ready for your review.

jdhoffa: I'm still getting plenty of test failures when I run the tests locally. This is with the dev version of `r2dii.*`

maurolepore: Thanks! Good that I now see the failures on gh-actions. I'll check them out.

jdhoffa:
So it seems like the most recent failing tests (on devel) are coming from the most recent commit to `r2dii.data`: 2DegreesInvesting/r2dii.data@4d83cc1
This was surprising to me, as I expected only `region_isos` to change from this commit, however `region_isos_demo` changed as well. Since `region_isos_demo` is used in most tests, this will cause small (sometimes large) differences in the regression values. 

Upon closer inspection, I can see that `region_isos_demo` is sourced using the actual values of `region_isos`: https://github.com/2DegreesInvesting/r2dii.data/blob/master/data-raw/region_isos_demo.R

As a result, `region_isos_demo` is not actually that stable. I think this is an issue, and it would make sense to freeze a stable version of `region_isos_demo` (either in `r2dii.analysis` or `r2dii.data`), again for the purposes of testing. This will be an easier job, since this data is passed explicitly to all functions as an argument. 

Let me know if you have questions, and hopefully we can fix this together tomorrow.

data-raw/sysdata.R
```diff
@@ -0,0 +1,39 @@
+# The `*_stable` datasets help create regression tests independent from changes
+# in r2dii.data which might cause meaningless errors when checking for problems
+# in reverse dependencies of r2dii.data. For example, in `loanbook_demo`, values
+# of the columns `sector` and `borderline` may change if the classification
+# system is updated (see #227).
```
maurolepore, line 5: I polished the text a bit.

data-raw/sysdata.R
```diff
@@ -0,0 +1,39 @@
...
+# Access older version of r2dii.data without changing current version
+temp_lib <- tempdir()
+fs::dir_create(temp_lib)
+
+withr::with_libpaths(temp_lib, {
+  remotes::install_version("r2dii.data", version = "0.1.4")
+  region_isos_demo <- r2dii.data::region_isos_demo
+  nace_classification <- r2dii.data::nace_classification
+  loanbook_demo <- r2dii.data::loanbook_demo
+})
```
maurolepore, line 20: I install the older version on a disposable library so we don't mess up with the current installation.

data-raw/sysdata.R
```diff
@@ -0,0 +1,39 @@
+# The `*_stable` datasets help create regression tests independent from changes
+# in r2dii.data which might cause meaningless errors when checking for problems
+# in reverse dependencies of r2dii.data. For example, in `loanbook_demo`, values
+# of the columns `sector` and `borderline` may change if the classification
+# system is updated (see #227).
+
+library(dplyr)
+library(fs)
+library(withr)
```
maurolepore, line 9: This is unecessary but mostly to clearly anounce which packages I'm using.

data-raw/sysdata.R
```diff
@@ -0,0 +1,39 @@
...
+
+loanbook_stable <- left_join(
+  loanbook_demo, nace_classification,
+  by = c(sector_classification_direct_loantaker = "code")
+)
+
+usethis::use_data(
+  region_isos_stable,
+  loanbook_stable,
+  internal = TRUE,
```
maurolepore, line 35: This results in R/sysdata.R so I renamed this file to data-raw/sysdata.R. All sysdata must be defined in the same call to `use_data()`, so this new name should make it easier to find the right place to add new sysdata.

maurolepore (approved):
Thanks @jdhoffa, your idea is a great one. I polished the script a bit. See my comments below. I'm happy to merge whenever you are ready. Locally, I see the checks pass cleanly both with release and devel versions of r2dii.data and r2dii.match.

maurolepore:
@jdhoffa, while we have this context in the head,  do you see an opportunity to reuse this idea to make r2dii.match more robust too?

data-raw/sysdata.R
jdhoffa: makes sense to me!

jdhoffa:
Ya let's think about it. Luckily `region_isos` doesn't play a role in `match_name()` at all, but could make sense to add a `loanbook_stable` object.

PR #231, opened 2020-11-27
R CMD check with devel version of other r2dii packages

maurolepore:
We can now reproduce problems in reverse dependencies that CRAN detected but not us, for example: https://github.com/2DegreesInvesting/r2dii.analysis/actions/runs/387692571

### DIFF
--- a/.github/workflows/R-CMD-check_r2dii-devel.yaml
+++ b/.github/workflows/R-CMD-check_r2dii-devel.yaml
@@ -79,8 +79,8 @@ jobs:
 
       - name: Install latest changes to reverse dependencies we maintain
         run: |
-          remotes::install_github("r2dii.data")
-          remotes::install_github("r2dii.match")
+          remotes::install_github("2DegreesInvesting/r2dii.data")
+          remotes::install_github("2DegreesInvesting/r2dii.match")
         shell: Rscript {0}
 
       - name: Session info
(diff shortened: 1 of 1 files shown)


.github/workflows/R-CMD-check_r2dii-devel.yaml
```diff
@@ -0,0 +1,109 @@
...
+        run: |
+          remotes::install_deps(dependencies = TRUE)
+          remotes::install_cran("rcmdcheck")
+        shell: Rscript {0}
+
+      - name: Install devel version of other r2dii packages
+        run: |
+          remotes::install_github("2DegreesInvesting/r2dii.data")
+          remotes::install_github("2DegreesInvesting/r2dii.match")
+        shell: Rscript {0}
```
maurolepore, line 84: This is the only bit that differs from the standard workflow. Here we install the development version of other r2dii packages before we run R CMD check on r2dii.analysis.

jdhoffa (approved):
Awesome, I back it!

jdhoffa:
I merged this even though the Checks added were failing. I think this makes sense since it reflects the truth that the current devel version of this package is failing. 

Also notice that https://github.com/2DegreesInvesting/r2dii.analysis/pull/230 will pass all `devel` checks, but fail all `release` checks, which seems like not the right solution. Lets discuss and try to resolve today if possible.

PR #242, opened 2020-12-04
Fix unweighted company

### LINKED ISSUE #239: Company level results projected error
The company-level results 

Company_results <- target_market_share(validated_matches, ald, scen, by_company = TRUE, weight_production = FALSE)

does not equal the projected (unweighted) values in the ALD for a given year, technology and company when there are multiple loans to the same company. 
If there is only one loan to the company in the loan book then the company results "projected" = the ALD production as expected.
 
This makes me think that despite the "weighted_production = FLASE" it is still weighting the results.

Loan book with 1 loan to each company - company-level results as expected 
[mock_lbk_1_loan.xlsx](https://github.com/2DegreesInvesting/r2dii.analysis/files/5623048/mock_lbk_1_loan.xlsx)

Loan book with multiple loans to each company - company-level results incorrect 
[mock_lbk_multi_loan.xlsx](https://github.com/2DegreesInvesting/r2dii.analysis/files/5623049/mock_lbk_multi_loan.xlsx)

jdhoffa:
`summarize_unweighted_production` now drops the crucial column `id_loan` and selects distinct values, to avoid erroneous double-counting when calculating the `unweighted` company production. 

Closes #239

### DIFF
--- a/tests/testthat/test-target_market_share.R
+++ b/tests/testthat/test-target_market_share.R
@@ -222,7 +222,7 @@ test_that("with known input outputs as expected, at company level", {
 
   expect_equal(
     out_target$production,
-    c(20, 180, 47.2, 305.8, 60, 190, 36, 114)
+    c(10, 90, 23.6, 152.9, 30, 95, 18, 57)
   )
 })
 
--- a/tests/testthat/test-target_market_share.R
+++ b/tests/testthat/test-target_market_share.R
@@ -521,3 +521,26 @@ test_that("w/ technology in ald but not loanbook, outputs all techs (#235)", {
   expect_equal(corporate_economy$ice$technology_share, 0.25)
   expect_equal(corporate_economy$electric$technology_share, 0.75)
 })
+
+test_that("w/ unweighted company flags & multi loans, outputs correctly (#239)", {
+
+  matched <- fake_matched(id_loan = c("L1", "L2"))
+
+  ald <- fake_ald()
+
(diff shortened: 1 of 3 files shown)


tests/testthat/test-target_market_share.R
```diff
@@ -521,3 +521,24 @@ test_that("w/ technology in ald but not loanbook, outputs all techs (#235)", {
   expect_equal(corporate_economy$ice$technology_share, 0.25)
   expect_equal(corporate_economy$electric$technology_share, 0.75)
 })
+
+test_that("w/ unweighted company flags & multi loans, outputs correctly (#239)", {
+  matched <- fake_matched(id_loan = c("L1", "L2"))
+
+  ald <- fake_ald()
+
+  scenario <- fake_scenario()
```
maurolepore, line 530: Nit: I think these three statements belong to the same "initialization paragraph" (a common pattern in tests is to divde the code in three paragraphs -- initialization, computation, expectation). I would use not white line between them.

maurolepore (approved):
LGTM

tests/testthat/test-target_market_share.R
jdhoffa: sounds good and agreed! do you have a resource that explains the pattern of test division?

maurolepore: Pretty sure it's in "Clean code". I have a hard copy here ... I'll see if I can find it.

PR #243, opened 2020-12-04
Reweight `technology_share` prior to output. 

### LINKED ISSUE #218: `technology_share` output should always sum to 1
Regardless of if `by_company = T` or `F`, the the sum of `technology_share` across sector/technology should always sum to 1.

jdhoffa:
Closes #218

This is not the most elegant solution, as I am simply re-weighting the `technology_share`, after it is calculated and summarized, which sorta feels like cheating, but I can't think of something more clever. 

The issue is that, for example, if we have company level results, we may have a table like: 
``` r
tibble::tribble(
  ~name_company, ~technology, ~technology_share,
            "a",  "electric",               0.5,
            "a",       "ice",               0.5,
            "b",  "electric",              0.25,
            "b",       "ice",              0.75
  )
#> # A tibble: 4 x 3
#>   name_company technology technology_share
#>   <chr>        <chr>                 <dbl>
#> 1 a            electric               0.5 
#> 2 a            ice                    0.5 
#> 3 b            electric               0.25
#> 4 b            ice                    0.75
```

<sup>Created on 2020-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

and if we then want to aggregate to portfolio level, a simple sum would result in: 
``` r
tibble::tribble(
  ~technology, ~technology_share,
   "electric",             0.375,
        "ice",             0.625
  )
#> # A tibble: 2 x 2
#>   technology technology_share
#>   <chr>                 <dbl>
#> 1 electric              0.375
#> 2 ice                   0.625
```

<sup>Created on 2020-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

which is basically a weighted sum of those. This can be achieved by simply aggregating the usual share, and then just re-normalizing (which is what I've done here). 

My brain is kinda mush right now, so if you can think of a more clever way to do this than what is in this PR, then by all means, but this satisfies the test condition for now. 

Closes #218

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -380,7 +380,7 @@ reweight_technology_share <- function(data, ...){
   data %>%
     group_by(...) %>%
     mutate(
-      .x = weighted_technology_share,
+      .x = .data$weighted_technology_share,
       weighted_technology_share = .data$.x/sum(.data$.x),
       .x = NULL
     ) %>%
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -376,15 +376,13 @@ empty_target_market_share_output <- function() {
   )
 }
 
-reweight_technology_share <- function(data, ...){
+reweight_technology_share <- function(data, ...) {
   data %>%
     group_by(...) %>%
     mutate(
       .x = .data$weighted_technology_share,
-      weighted_technology_share = .data$.x/sum(.data$.x),
+      weighted_technology_share = .data$.x / sum(.data$.x),
       .x = NULL
     ) %>%
     ungroup()
-
-
 }
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -107,25 +107,23 @@ target_market_share <- function(data,
     by_company
   )
 
-  data <- data %>%
-    join_ald_scenario(ald, scenario, region_isos) %>%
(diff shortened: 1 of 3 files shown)


R/target_market_share.R
```diff
@@ -367,3 +375,16 @@ empty_target_market_share_output <- function() {
     technology_share = numeric(0)
   )
 }
+
+reweight_technology_share <- function(data, ...){
```
maurolepore, line 379: Can you please run `use_tidy_style()`?

tests/testthat/test-target_market_share.R
```diff
@@ -541,3 +541,26 @@ test_that("w/ unweighted company flags & multi loans, outputs correctly (#239)",
 
   expect_equal(projected$production, ald$production)
 })
+
+test_that("w/ multiple loans to same company, `technology_share` sums to one (#218)", {
+
```
maurolepore, line 546: /style

maurolepore (approved):
LGTM. Please style (maybe comment with `/style`) and merge.
Nice code and reprex!

maurolepore (approved):
/style did the trick so I'll go ahead and merge. Thanks!

PR #263, opened 2021-01-06
Aggregate all loans to the same company in `target_market_share`

jdhoffa:
As the first step of `target_market_share`, I now aggregate by `id_loan` to the same company on the input data. We likely could have done this much earlier, I don't recall now why I didn't... but all tests are passing and everything looks good so I think it should be ok!

Closes #262

Issue #262, opened 2021-01-06
`technology_share` values seem off


jdhoffa:
See reprex. Expected values were manually calculated in excel. 

---

``` r
library(r2dii.data)
library(r2dii.analysis)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

matched <- tibble::tribble(
  ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,             ~level, ~score, ~sector,   ~name_ald, ~sector_ald,
  "L1",                  1000L,                           "USD",                   1000L,                            "USD",    "DL1", "direct_loantaker",     1L, "power", "company a",     "power",
  "L2",                 15000L,                           "USD",                  15000L,                            "USD",    "DL2", "direct_loantaker",     1L, "power", "company b",     "power",
  "L3",                  1200L,                           "USD",                   1200L,                            "USD",    "DL1", "direct_loantaker",     1L, "power", "company a",     "power"
)

ald <- tibble::tribble(
  ~name_company, ~sector,     ~technology, ~year, ~production, ~emission_factor, ~plant_location, ~is_ultimate_owner,
  "company a", "power",       "coalcap", 2020L,        100L,              0.5,            "US",               TRUE,
  "company a", "power",        "gascap", 2020L,        200L,              0.2,            "US",               TRUE,
  "company a", "power",      "hydrocap", 2020L,        300L,                0,            "US",               TRUE,
  "company a", "power",    "nuclearcap", 2020L,        100L,                0,            "US",               TRUE,
  "company a", "power",        "oilcap", 2020L,        100L,              0.3,            "US",               TRUE,
  "company a", "power", "renewablescap", 2020L,        200L,                0,            "US",               TRUE,
  "company b", "power",       "coalcap", 2020L,        500L,              0.5,            "US",               TRUE,
  "company b", "power",        "gascap", 2020L,          0L,              0.2,            "US",               TRUE,
  "company b", "power",      "hydrocap", 2020L,          0L,                0,            "US",               TRUE,
  "company b", "power",    "nuclearcap", 2020L,        300L,                0,            "US",               TRUE,
[...]

PR #268, opened 2021-01-18
`target_market_share` errors if input `data` has unexpected columns

jdhoffa:
Closes #267
Thanks for flagging the issue @georgeharris2deg 

I elected to go the more restrictive route since there is some internal aggregations and groupings happening, it's impossible to know arbitrarily how to handle any unexpected column. 

This option will prevent the user from even being able to input unexpected columns, even if they want to. Happy to revisit the decision, but I think it's safest for now.

PR #270, opened 2021-01-19
Initially aggregate loans by appropriate groups. 

### LINKED ISSUE #265: Technology share values persist being off due to aggregation by name_ald
Function `add_technology_share()` from module _summarize_weighted_production.R_ groups data by _name_ald_ and this generates the technology mix or technology share for each company -- it will later be multiplied by the loan weights and summed over to give the final (weighted) technology share.

`add_technology_share <- function(data) {
  data %>%
    group_by(.data$sector_ald, .data$year, .data$scenario, .data$name_ald) %>%
    mutate(technology_share = .data$production / sum(.data$production)) %>%
    group_by(!!!dplyr::groups(data))
}`

The issue arises when several _name_direct_loantakers_ are matched to one unique _name_ald_ (this happens when matches are performed at the _ultimate_parent_ level). Function `add_technology_share()` will aggregate all repeated instances of _name_ald_ (there will be as many as the number of direct loantakers assigned to the ALD company) and sum production, yielding a bigger total production and thus a lower technology share than the real one. 

**Substituting _name_ald_ by _name_direct_loantaker_** in this function **solves the issue**. I'm attaching a simplified example for visualization purposes.
image

jdhoffa:
I was hasty in writing #263, in which I added an aggregation step to aggregate similar loans to different companies, if these loans mapped to the same ALD company. I used a `group_by_at(...)`, and `setdiff()` to aggregate by all but a few columns, which was lazy of me, and didn't think about the downstream issues of grouping by a set of columns that should in fact map to the same information.  

Here, I have made the aggregation step more specific, by grouping only by the relevant columns:
``` r
group_by(
      .data$level,
      .data$loan_size_outstanding_currency,
      .data$loan_size_credit_limit_currency,
      .data$name_ald,
      .data$sector_ald
      ) 
```

Relates to #262 
Closes #265

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -94,7 +94,7 @@ target_market_share <- function(data,
   }
 
   data <- ungroup(warn_grouped(data, "Ungrouping input data."))
-
+  check_unexpected_columns(data)
   data <- aggregate_by_loan_id(data)
 
   crucial_scenario <- c("scenario", "tmsr", "smsp")
@@ -390,7 +390,13 @@ reweight_technology_share <- function(data, ...) {
 }
 
 aggregate_by_loan_id <- function(data) {
-  aggregate_columns <- c("id_loan", "loan_size_outstanding", "loan_size_credit_limit")
+
+  aggregate_columns <- c(
+    "id_loan",
+    "loan_size_outstanding",
+    "loan_size_credit_limit"
+    )
+
 
   data %>%
     dplyr::group_by_at(setdiff(names(data), aggregate_columns)) %>%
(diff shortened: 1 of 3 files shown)


R/target_market_share.R
```diff
@@ -390,14 +390,65 @@ reweight_technology_share <- function(data, ...) {
...
+      ) %>%
     summarize(
       id_loan = first(.data$id_loan),
       loan_size_outstanding = sum(.data$loan_size_outstanding),
       loan_size_credit_limit = sum(.data$loan_size_credit_limit)
     ) %>%
     ungroup()
 }
+
+check_unexpected_columns <- function(data) {
```
maurolepore, line 410: I saw this in another open PR. Do they depend on each other? Do we need to make that explicit in the first line of the opening PR comment?

R/target_market_share.R
```diff
@@ -390,14 +390,65 @@ reweight_technology_share <- function(data, ...) {
...
-  aggregate_columns <- c("id_loan", "loan_size_outstanding", "loan_size_credit_limit")
 
   data %>%
-    dplyr::group_by_at(setdiff(names(data), aggregate_columns)) %>%
+    group_by(
+      .data$level,
+      .data$loan_size_outstanding_currency,
+      .data$loan_size_credit_limit_currency,
+      .data$name_ald,
+      .data$sector_ald
```
maurolepore, line 400: We don't have an object holding this, right? Maybe some `crucial_market_share_cols` or something like that sitting somewhere else? Just wondering.

maurolepore (approved):
LGTM but WARNING: This PR seems to depend on another open PR. Please remember to handle that dependency.

R/target_market_share.R
jdhoffa: Ugh woops ya, i made a mistake and branched off the open PR. Will handle quickly today

jdhoffa:
I'm going to close this, and open a new PR request that doesn't depend on #268

R/target_market_share.R
jdhoffa: nay we do not afaik

Issue #277, opened 2021-02-11
Initial value of `technology_share` differs between `projected` and `target_*`


jdhoffa:
The `technology_share` should be identical between the `projected` and `target_*` values at the start year of the analysis. In the reprex below, we can see that this is not the case. 

I have a hunch as to why this is, I believe the `target` technology_share is being calculated the old way (dividing `weighted_production` by the sum of `weighted_production`). Need to explore how to solve this. 

Thanks as always @georgeharris2deg 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(r2dii.data)
library(r2dii.analysis)

matched <- tibble::tribble(
  ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,            ~level, ~score,      ~sector,   ~name_ald,  ~sector_ald,
  "L1",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive", "company a", "automotive",
  "L2",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive", "company b", "automotive"
)

ald <- tibble::tribble(
  ~name_company,      ~sector, ~technology, ~year, ~production, ~emission_factor, ~plant_location, ~is_ultimate_owner,
  "company a", "automotive",       "ice",  2020,         100,                1,            "BF",               TRUE,
  "company b", "automotive",       "ice",  2020,           1,                1,            "BF",               TRUE,
  "company a", "automotive",  "electric",  2020,         100,                1,            "BF",               TRUE,
  "company b", "automotive",  "electric",  2020,           3,                1,            "BF",               TRUE
)

scenario <- tibble::tribble(
  ~scenario,      ~sector, ~technology,  ~region, ~year, ~tmsr, ~smsp, ~scenario_source,
  "sds", "automotive",       "ice", "global",  2020,     1,     0,      "demo_2020",
  "sds", "automotive",  "electric", "global",  2020,     1,     0,      "demo_2020"
)

out <- target_market_share(matched, ald, scenario, region_isos_demo)

out %>% 
  filter(
    metric %in% c("projected", "target_sds"),
[...]

PR #280, opened 2021-02-18
Refactor `target_market_share` (#277)

jdhoffa:
This PR is strictly refactoring, and should not change the expected functionality of the code whatsoever. Most of the refactoring occurs in `target_market_share`, although I briefly touched `summarize_weighted_production` as I noticed something weird about the interface of the functions. 

The main change can be thought of as follows. Previously, the function `target_market_share()` worked something like this: 
- Join input datasets together
- Add a metric AND summarize (either to company level, or to portfolio level)
- Add targets 
- Tidy-up a bit and output

Now, the function works more like this: 
- Join input datasets together
- Add a metric AND summarize (strictly to company level)
- Add targets 
- Maybe aggregate to portfolio level
- Tidy-up a bit and output

This will make it easier to address the bugfix necessary in #277. 
Relates to #277

### DIFF
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -70,7 +70,7 @@ summarize_unweighted_production <- function(data, ...) {
       .data$loan_size_outstanding
     )) %>%
     distinct() %>%
-    group_by(...) %>%
+    group_by(.data$sector_ald, .data$technology, .data$year, ...) %>%
     summarize(weighted_production = sum(.data$production), .groups = "keep") %>%
     ungroup(.data$technology) %>%
     mutate(weighted_technology_share = .data$weighted_production / sum(.data$weighted_production)) %>%
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -164,52 +164,23 @@ target_market_share <- function(data,
     return(empty_target_market_share_output())
   }
 
-  target_groups <- c("sector_ald", "scenario", "year", "region", "name_ald")
-
-  initial_sector_summaries <- data %>%
-    group_by(!!!rlang::syms(target_groups)) %>%
-    summarize(sector_weighted_production = sum(.data$weighted_production)) %>%
-    arrange(.data$year) %>%
-    group_by(
-      .data$sector_ald,
-      .data$scenario,
-      .data$region,
-      .data$name_ald
-      ) %>%
-    filter(row_number() == 1L) %>%
-    rename(
-      initial_sector_production = .data$sector_weighted_production
-    ) %>%
-    select(-.data$year)
-
-  initial_technology_summaries <- data %>%
-    group_by(!!!rlang::syms(c(target_groups, "technology"))) %>%
-    summarize(
-      technology_weighted_production = sum(.data$weighted_production)
-    ) %>%
-    arrange(.data$year) %>%
-    group_by(
-      .data$sector_ald,
-      .data$technology,
-      .data$scenario,
-      .data$region,
-      .data$name_ald
-    ) %>%
-    filter(row_number() == 1L) %>%
-    rename(
-      initial_technology_production = .data$technology_weighted_production
-    ) %>%
-    select(-.data$year)
+  target_groups <- c("sector_ald", "scenario", "region", "name_ald")
 
   data <- data %>%
-    left_join(
-      initial_sector_summaries,
-      by = c("sector_ald", "scenario", "region", "name_ald")
-    ) %>%
-    left_join(
-      initial_technology_summaries,
-      by = c("sector_ald", "scenario", "region", "technology", "name_ald")
-    ) %>%
+    group_by(!!!rlang::syms(c(target_groups, "year"))) %>%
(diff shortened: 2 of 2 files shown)


R/summarize_weighted_production.R
```diff
@@ -70,7 +70,7 @@ summarize_unweighted_production <- function(data, ...) {
       .data$loan_size_outstanding
     )) %>%
     distinct() %>%
-    group_by(...) %>%
+    group_by(.data$sector_ald, .data$technology, .data$year, ...) %>%
```
jdhoffa, line 73: This change was just to align the interface of `summarize_weighted_production` and `summarize_unweighted_production`. Both functions now automatically group by `sector_ald, technology and year`, regardless of input.

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
   )
 
   check_valid_columns(data, valid_columns)
-  data <- aggregate_by_loan_id(data)
```
jdhoffa, line 130: I moved this down, so that the checks are separated from the "actions"

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
...
 
   crucial_scenario <- c("scenario", "tmsr", "smsp")
   check_crucial_names(scenario, crucial_scenario)
   check_crucial_names(ald, "is_ultimate_owner")
   walk_(crucial_scenario, ~ check_no_value_is_missing(scenario, .x))
 
   green_or_brown <- r2dii.data::green_or_brown
+  tmsr_or_smsp <- tmsr_or_smsp()
```
jdhoffa, line 137: This dataset is defined by a function, and called directly later in the script. I thought it would make sense to specify explicitly where the dataset is coming from when we import the other necessary datasets.

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
...
 
-  summary_groups <- maybe_add_name_ald(
-    c("scenario", "tmsr", "smsp", "region", "scenario_source"),
-    by_company
```
jdhoffa, line 141: `maybe_add_name_ald` is a function that adds (or not) the grouping variable `name_ald` to a list. In this PR, I have removed calls to this and similar functions many times, and opted to assume that `name_ald` should always be present, until the end of the script where a final summary is calculated.

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
...
-      use_credit_limit = use_credit_limit
+      !!!rlang::syms(summary_groups)
     )
+
   } else {
     data <- summarize_unweighted_production(
       data,
-      .data$sector_ald,
-      .data$technology,
-      .data$year,
```
jdhoffa, line 157: these are now internally grouped in the body of the function `summarize_unweighted_production()`

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
...
-  reweighting_groups <- maybe_add_name_ald(
-    c("sector_ald", "region", "scenario", "scenario_source", "year"),
-    by_company
-  )
-
-  data <- reweight_technology_share(
-    data,
-    !!!rlang::syms(reweighting_groups)
-  )
-
```
jdhoffa, line 171: this happens later

R/target_market_share.R
```diff
@@ -127,102 +127,62 @@ target_market_share <- function(data,
...
-        by_company = by_company
-      )
-    ) %>%
-    left_join(
-      initial_technology_summaries,
-      by = maybe_add_name_ald(
-        c("sector_ald", "scenario", "region", "technology"),
-        by_company = by_company
-      )
-    ) %>%
```
jdhoffa, line 225: Before: this chunk was done in a really awkward way. I calculate two data-frames from `data`: `inital_sector_summaries` and `initial_technology_summaries`, and then `join` them back into the original `data`

Now: This is done more elegantly with a couple calls to `group_by` and `summarize`

R/target_market_share.R
```diff
@@ -255,11 +215,37 @@ target_market_share <- function(data,
...
+    )
+
+    data <- data %>%
+      group_by(!!!rlang::syms(aggregate_company_groups)) %>%
+      summarize(
+        weighted_production = sum(.data$weighted_production),
+        weighted_production_target = sum(.data$weighted_production_target),
+        weighted_technology_share = sum(.data$weighted_technology_share)
+      )
+  }
```
jdhoffa, line 236: this was initially done directly by `summarize_weighted_production`, but by adding `name_ald` to the summary groups, we need to actually summarize again at the end, and potentially aggregate to portfolio level. I think separating the one summary into two distinct summaries, which aim to achieve different things, is better

maurolepore (approved):
LGTM

R/target_market_share.R
jdhoffa: Sure that makes sense. My hope was to elucidate that data was necessary that doesn't necessarily come from an argument, and to clearly define it, but I see the point. I'll move them closer to where they are used.

PR #281, opened 2021-02-19
Refactor `summarize_weighted_production`

jdhoffa:
Here comes a pretty hefty refactoring. The bulk of my struggles come from the fact that the function `summarize_weighted_production` is: 

1. An exported function
2. Poorly named (It summarizes weighted `production` AND `technology_share`

While playing within the boundaries of keeping the function doing what it says it does, I am trying to extract some of the common functionalities of `summarize_weighted_production`, `summarize_weighted_percent_change` and `summarize_weighted_emission_factor`, and allow them to be a little more flexible. 

What I will end up needing to fix the bug in #277 is something along the lines of `summarize_weighted_production_and_technology_share_and_target_and_target_technology_share` (which is obviously a joke, but it highlights the fact that we should have a more flexible summarize_ type function.)

The `summarize_weighted_metric` function unfortunately is not flexible enough to suit this purpose. The `.f` argument only allows a single input function, and beyond this, each metric requires some little differences to be adjusted on the input data. (Don't worry if none of this makes sense, I'll explain in comments in the code, and it's also more so I remember what I was working on when I start again on Monday lol). 

I leave it in draft mode, because I'm not quite finished with it, but would appreciate some help @maurolepore if it's possible to find the time to do so (no problem if not). 

Relates to #277


tests/testthat/test-summarize_weighted_production.R
```diff
@@ -306,7 +306,7 @@ test_that("with bad but unused loan_size_column is error free", {
 
   bad_unused <- rename(fake_master(), bad = "loan_size_outstanding")
   expect_error_free(
-    add_weighted_loan_percent_change(bad_unused, use_credit_limit = TRUE)
+    summarize_weighted_percent_change(bad_unused, use_credit_limit = TRUE)
```
jdhoffa, line 309: I believe these were typos, since the assertion above is referencing `summarize_weighted_percent_change`

tests/testthat/test-summarize_weighted_production.R
```diff
@@ -81,7 +81,7 @@ test_that("with bad but unused loan_size_column is error free", {
 
   bad_unused <- rename(fake_master(), bad = "loan_size_outstanding")
   expect_error_free(
-    add_weighted_loan_production(bad_unused, use_credit_limit = TRUE)
+    summarize_weighted_production(bad_unused, use_credit_limit = TRUE)
```
jdhoffa, line 84: I believe these were typos, since the assertion above is referencing `summarize_weighted_percent_change`

R/target_sda.R
```diff
@@ -224,25 +235,6 @@ maybe_add_name_ald <- function(data, by_company = FALSE) {
...
-      use_credit_limit = use_credit_limit
-    ) %>%
-    group_by(!!!rlang::syms(summary_groups)) %>%
-    summarize(
-      emission_factor_projected = sum(.data$weighted_loan_emission_factor)
-    ) %>%
-    ungroup() %>%
-    rename(sector = .data$sector_ald)
-}
-
```
jdhoffa, line 245: I moved this function to `summarize_weighted_production.R`

R/target_sda.R
```diff
@@ -130,14 +130,25 @@ target_sda <- function(data,
...
+      summarize_weighted_emission_factor(
+        "name_ald",
+        use_credit_limit = use_credit_limit
+      )
+  } else {
+    data <- data %>%
+      summarize_weighted_emission_factor(
+        use_credit_limit = use_credit_limit
+      )
+    }
```
jdhoffa, line 146: I moved `summarize_weighted_emission_factor` to `./R/summarize_weighted_production.R`, and adjusted it's interface to reflect that of the other `summarize_weighted_*` functions

R/target_market_share.R
```diff
@@ -133,9 +133,13 @@ target_market_share <- function(data,
...
   walk_(crucial_scenario, ~ check_no_value_is_missing(scenario, .x))
 
+  data <- aggregate_by_loan_id(data)
 
+  data <- join_ald_scenario(data, ald, scenario, region_isos)
 
-  data <- aggregate_by_loan_id(data)
+  if (nrow(data) == 0) {
+    return(empty_target_market_share_output())
+  }
```
jdhoffa, line 142: Moving things around.

tests/testthat/test-target_sda.R
```diff
@@ -20,6 +20,25 @@ test_that("with fake data outputs known value", {
...
+      emission_factor = c(1, 2, 3)
+    ),
+    co2_intensity_scenario = fake_co2_scenario(
+      year = c(2020, 2050),
+      emission_factor = c(0.6, 0.2)
+    ),
+    by_company = TRUE
+  )
+
+  expect_known_value(out_company, "ref-target_sda_company", update = TRUE)
```
jdhoffa, line 41: Some of the refactoring caused the checks to fail when generating the examples. I added a test to help with this.

maurolepore: Is that failure the reason to use `expect_known_value()` instead of `expect_snapshot()`?

jdhoffa: No, I only use `expect_known_value()` as this is what the rest of the package seems to be using still. Maybe in a separate refactoring we can switch everything to `expet_snapshot()`?

maurolepore: Thanks! The issue with separate refactoring is that it rarely happens. "Cleaning as you go" tends to keep the house cleaner than "I'll do it on Saturday"

jdhoffa: Sure, I guess my hope was to keep this PR relatively focused (I even seem to have already failed that). Changing the `expect_known_value()` to `expect_snapshot()` in every test that it occurs in, seems like it is completely unrelated to this PR and it's purpose. 

Also, I was very hesitant to touch the tests at all in this PR, as the purpose is refactoring and I wanted to ensure the output was consistent with before. This was (hopefully) the only place that I touched a test, as I was actually adding a test where we were lacking coverage.

maurolepore (approved):
LGTM.

PR #282, opened 2021-02-22
Dont group_by `level` in `target_market_share()`

### LINKED ISSUE #265: Technology share values persist being off due to aggregation by name_ald

jdhoffa:
The `level` column is only really used in the matching process for calculating coverage statistics. 
In the actual analysis, we should ignore this column. 

Maybe closes #265

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # r2dii.analysis (development version)
 
+* `target_market_share()` now correctly outputs `technology_share` with
+  multiple loans at different `level` to the same company (@ab-bbva #265).
+
 # r2dii.analysis 0.1.5
 
 * `target_market_share()` now errors if input `data` has an unexpected column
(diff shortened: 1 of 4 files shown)


maurolepore (commented):
LGTM. Do we need to add an entry to NEWS thanking the person who reported the issue?

maurolepore (approved):

PR #287, opened 2021-04-06
Update DESCRIPTION's title and description

maurolepore:
As I'm touching DESCRIPTION I realize this package lacks the helpfile for `?r2dii.analysis`. For consistency I add it here. Although I'm happy to move that commit to a separate PR, I think it's somewhat appropriate here because the similar PRs in r2dii.data and r2dii.match touch not only DESCRIPTION but also man/r2dii.*-package.Rd.

I'm not adding an entry to NEWS because I'm not entirely sure it's appropriate as this change is exposed to users in a too subtle way -- but I'm happy to add it if @jdhoffa thinks it's useful.


R/r2dii.analysis-package.R
```diff
@@ -0,0 +1,8 @@
+#' @keywords internal
+"_PACKAGE"
+
+# The following block is used by usethis to automatically manage
+# roxygen namespace tags. Modify with care!
+## usethis namespace: start
+## usethis namespace: end
+NULL
```
jdhoffa, line 8: LGTM, but what's the use of this?

jdhoffa (approved):

R/r2dii.analysis-package.R
maurolepore: Roxygen2 uses this file to automatically generate the helpfile you see when you run `?r2dii.analysis`. In turn, you generate this file with `usethis::use_package_doc()`.

When I hear of a cool new package and have no idea 
image

--

At the very bottom, that  helpfile has a convenient link to the entire index, which is a great way to see all features at a glance:

image

PR #289, opened 2021-04-28
Enable testing a single file (namespace pull())

maurolepore:


maurolepore (commented):
This is a safe change with no impact on observable bahavour. It improves the quality of life of the developers, by allowing them to run `test_file()` (usually shift + control + T) rather than the entire test suite.

PR #292, opened 2021-05-03
target_market_share() no longer returns internal columns

### LINKED ISSUE #291: `target_market_share(..., by_company = TRUE)` outputs unexpected columns
In particular, when comparing the output of `target_market_share(..., by_company = FALSE)` and `target_market_share(..., by_company = TRUE)`, I would expect to see only the added column `name_ald`. 

What I see in practice is three added columns: `name_ald`, and then the two unexpected columns `sector_weighted_production` and `technology_weighted_production`.

``` r
library(r2dii.data)
library(r2dii.match) 
library(r2dii.analysis)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 

matched <- match_name(loanbook_demo, ald_demo) %>%
  prioritize()

out_false <- target_market_share(
  matched,
  ald_demo,
  scenario_demo_2020,
  region_isos_demo,
  by_company = FALSE
)

out_true <- target_market_share(
  matched,
  ald_demo,
  scenario_demo_2020,
  region_isos_demo,
  by_company = TRUE
  )
#> Warning: You've supplied `by_company = TRUE` and `weight_production = TRUE`.
#> This will result in company-level results, weighted by the portfolio
#> loan size, which is rarely useful. Did you mean to set one of these
#> arguments to `FALSE`?

# similarly, as expected, if by_company = TRUE, the output has the same columns 
# regardless of if weight_production = T or F
setdiff(names(out_true), names(out_false))
#> [1] "name_ald"                       "sector_weighted_production"    
#> [3] "technology_weighted_production"

# I would expect the only added column to be `name_ald`
setdiff(names(out_true), names(out_false)) == "name_ald"
#> [1]  TRUE FALSE FALSE
```

<sup>Created on 2021-05-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

maurolepore:
Closes #291﻿

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -21,9 +21,9 @@
 #' outputting production, weighted by relative loan-size. Set to `FALSE` to
 #' output the unweighted production values.
 #'
-#' @return A tibble with the summarized columns `metric`, `production` and
-#' `technology_share`. If `by_company = TRUE`, the output will also have the
-#' column `name_ald`.
+#' @return A tibble including the summarized columns `metric`, `production` and
+#'   `technology_share`. If `by_company = TRUE`, the output will also have the
+#'   column `name_ald`.
 #' @export
 #'
 #' @family functions to calculate scenario targets
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -175,14 +175,16 @@ target_market_share <- function(data,
     mutate(sector_weighted_production = sum(.data$weighted_production)) %>%
     arrange(.data$year) %>%
     group_by(!!!rlang::syms(target_groups)) %>%
-    mutate(initial_sector_production = first(.data$sector_weighted_production))
+    mutate(initial_sector_production = first(.data$sector_weighted_production)) %>%
+    select(-.data$sector_weighted_production)
 
   data <- data %>%
     group_by(!!!rlang::syms(c(target_groups, "technology", "year"))) %>%
     mutate(technology_weighted_production = sum(.data$weighted_production)) %>%
     arrange(.data$year) %>%
     group_by(!!!rlang::syms(c(target_groups, "technology"))) %>%
-    mutate(initial_technology_production = first(.data$technology_weighted_production))
+    mutate(initial_technology_production = first(.data$technology_weighted_production)) %>%
+    select(-.data$technology_weighted_production)
 
   green_or_brown <- r2dii.data::green_or_brown
   tmsr_or_smsp <- tmsr_or_smsp()
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # r2dii.analysis (development version)
 
+* `target_market_share()` no longer outputs columms
+  `sector_weighted_production` or `technology_weighted_production`. Those
+  columns are internal so they shouldn't face users.
+
 # r2dii.analysis 0.1.6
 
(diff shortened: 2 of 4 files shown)


R/target_market_share.R
```diff
@@ -175,14 +175,16 @@ target_market_share <- function(data,
     mutate(sector_weighted_production = sum(.data$weighted_production)) %>%
     arrange(.data$year) %>%
     group_by(!!!rlang::syms(target_groups)) %>%
-    mutate(initial_sector_production = first(.data$sector_weighted_production))
+    mutate(initial_sector_production = first(.data$sector_weighted_production)) %>%
+    select(-.data$sector_weighted_production)
```
maurolepore, line 179: `sector_weighted_production` seems like an internal, temporary column, so I remove it ASAP.

R/target_market_share.R
```diff
@@ -175,14 +175,16 @@ target_market_share <- function(data,
...
 
   data <- data %>%
     group_by(!!!rlang::syms(c(target_groups, "technology", "year"))) %>%
     mutate(technology_weighted_production = sum(.data$weighted_production)) %>%
     arrange(.data$year) %>%
     group_by(!!!rlang::syms(c(target_groups, "technology"))) %>%
-    mutate(initial_technology_production = first(.data$technology_weighted_production))
+    mutate(initial_technology_production = first(.data$technology_weighted_production)) %>%
+    select(-.data$technology_weighted_production)
```
maurolepore, line 188: `technology_weighted_production` seems like an internal, temporary column, so I remove it ASAP.

R/target_market_share.R
```diff
@@ -21,9 +21,9 @@
 #' outputting production, weighted by relative loan-size. Set to `FALSE` to
 #' output the unweighted production values.
 #'
-#' @return A tibble with the summarized columns `metric`, `production` and
-#' `technology_share`. If `by_company = TRUE`, the output will also have the
-#' column `name_ald`.
+#' @return A tibble including the summarized columns `metric`, `production` and
+#'   `technology_share`. If `by_company = TRUE`, the output will also have the
+#'   column `name_ald`.
```
maurolepore, line 26: I think the original paragraph is unclear. I interpret it as meaning that the input and output have the same names except that the output has the new columns `metric`, `production`, `technology_share`, and `name_ald` (if `by_company = TRUE`) -- which is not true:

``` r
devtools::load_all()
#> ℹ Loading r2dii.analysis

input <- fake_matched()
output <- target_market_share(
  input,
  ald = fake_ald(),
  scenario = fake_scenario(),
  region_isos = region_isos_stable
)

setdiff(names(input), names(output))
#>  [1] "id_loan"                         "loan_size_outstanding"          
#>  [3] "loan_size_outstanding_currency"  "loan_size_credit_limit"         
#>  [5] "loan_size_credit_limit_currency" "id_2dii"                        
#>  [7] "level"                           "score"                          
#>  [9] "name_ald"                        "sector_ald"
```

So I added "including" to be more vague and more likely correct. If you can think of a better wording -- one that better reflects the returned value -- then please go ahead and edit.

jdhoffa (approved):
LGTM!

R/target_market_share.R
jdhoffa: I think this looks pretty good to me!

PR #294, opened 2021-05-10
`technology_share` for `target_*` is now properly calculated and weighted

### LINKED ISSUE #277: Initial value of `technology_share` differs between `projected` and `target_*`

jdhoffa:
The `target_*` values of `technology_share` should be calculated as follows (in line with the methodology): 
1 - Calculate the `production_target` for each company (without weighting) 
2 - Calculate the `technology_share_target` for each company (without weighting)
3 - Apply the portfolio_weight to each of `production`, `technology_share`, `production_target` and `technology_share_target`

This has been achieved. 
@maurolepore this is an MVP solution, as this has been an open bug for some time. Please review (and note that I recognize that some of my solutions are not so elegant). 

Also @georgeharris2deg if you could please have a look at this it would be much appreciated. 

I would rather we get the solution out the door, and we can refactor later. 

Closes #277

### DIFF
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -53,7 +53,24 @@
 #'
 #' summarize_weighted_percent_change(master, use_credit_limit = TRUE)
 summarize_weighted_production <- function(data, ..., use_credit_limit = FALSE) {
-  stopifnot(is.data.frame(data))
+  stopifnot(
+    is.data.frame(data),
+    is.logical(use_credit_limit)
+  )
+
+  crucial <- c(
+    "id_loan",
+    "loan_size_outstanding",
+    "loan_size_outstanding_currency",
+    "loan_size_credit_limit",
+    "loan_size_credit_limit_currency",
+    "sector_ald",
+    "name_ald",
+    "year",
+    "scenario"
+  )
+
+  check_crucial_names(data, crucial)
 
   data %>%
     ungroup() %>%
@@ -121,7 +138,6 @@ summarize_weighted_emission_factor <- function(data, ..., use_credit_limit = FAL
 calculate_weighted_loan_metric <- function(data, metric) {
   crucial <- c(metric, "loan_weight")
 
-  check_crucial_names(data, crucial)
   walk_(crucial, ~ check_no_value_is_missing(data, .x))
 
   data %>%
@@ -144,7 +160,6 @@ add_loan_weight <- function(data, use_credit_limit) {
     "id_loan", "sector_ald", "year", loan_size, currency
   )
 
-  check_crucial_names(data, crucial)
   walk_(crucial, ~ check_no_value_is_missing(data, .x))
 
   old_groups <- dplyr::groups(data)
@@ -166,11 +181,6 @@ add_loan_weight <- function(data, use_credit_limit) {
 }
 
 add_percent_change <- function(data) {
-  crucial <- c("production", "sector_ald", "year", "technology", "scenario")
-
-  check_crucial_names(data, crucial)
-  walk_(crucial, ~ check_no_value_is_missing(data, .x))
-
   check_zero_initial_production(data)
 
   green_or_brown <- r2dii.data::green_or_brown
@@ -200,11 +210,6 @@ add_percent_change <- function(data) {
 }
 
 add_technology_share <- function(data) {
-  crucial <- c("production", "sector_ald", "year", "technology")
-
-  check_crucial_names(data, crucial)
-  walk_(crucial, ~ check_no_value_is_missing(data, .x))
-
   data %>%
     group_by(.data$sector_ald, .data$year, .data$scenario, .data$name_ald) %>%
     mutate(technology_share = .data$production / sum(.data$production)) %>%
@@ -212,11 +217,6 @@ add_technology_share <- function(data) {
 }
 
 add_technology_share_target <- function(data) {
-  crucial <- c("production_target", "sector_ald", "year", "technology")
-
-  check_crucial_names(data, crucial)
-  walk_(crucial, ~ check_no_value_is_missing(data, .x))
-
   data %>%
     group_by(.data$sector_ald, .data$year, .data$scenario, .data$name_ald) %>%
     mutate(technology_share_target = .data$production_target / sum(.data$production_target)) %>%
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -53,24 +53,7 @@
 #'
 #' summarize_weighted_percent_change(master, use_credit_limit = TRUE)
 summarize_weighted_production <- function(data, ..., use_credit_limit = FALSE) {
-  stopifnot(
-    is.data.frame(data),
-    is.logical(use_credit_limit)
-  )
-
-  crucial <- c(
-    "id_loan",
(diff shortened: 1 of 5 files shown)


R/summarize_weighted_production.R
```diff
@@ -211,6 +211,18 @@ add_technology_share <- function(data) {
...
 
+add_technology_share_target <- function(data) {
+  crucial <- c("production_target", "sector_ald", "year", "technology")
+
+  check_crucial_names(data, crucial)
+  walk_(crucial, ~ check_no_value_is_missing(data, .x))
+
+  data %>%
+    group_by(.data$sector_ald, .data$year, .data$scenario, .data$name_ald) %>%
+    mutate(technology_share_target = .data$production_target / sum(.data$production_target)) %>%
```
maurolepore, line 222: Do we also need to check for crucial names "scenario" and "name_ald"? 

It would be more readable if `crucial` holds the names in the order they show up in the code, or in alphabetical order so they are easier to lookup.

R/target_market_share.R
```diff
@@ -220,6 +221,60 @@ target_market_share <- function(data,
...
+        .data$sector_ald,
+        .data$technology,
+        .data$year,
+        !!!rlang::syms(summary_groups)
+      ) %>%
+      summarize(
+        weighted_production = sum(.data$weighted_loan_production),
+        weighted_technology_share = sum(.data$weighted_loan_technology_share),
+        weighted_production_target = sum(.data$weighted_loan_production_target),
+        weighted_technology_share_target = sum(.data$weighted_loan_technology_share_target)
```
maurolepore, line 251: It's sometimes useful to call `sum()` with `rm.na = TRUE`. Likely you took this into account but just checking you didn't forget.

maurolepore (approved):
LGTM. I have a few questions you may consider and merge.

R/summarize_weighted_production.R
jdhoffa: I'm actually not sure what the protocol is here. These are internal only functions, and used in dplyr chains on data that has already had it's crucial columns checked. 

Would it make sense to remove the crucial check altogether?

maurolepore: Oh sure! You are right!  I think it's best to check names only once at the top level.

maurolepore:
Thanks @jdhoffa for your comments. I see a bunch of things we should follow up on and I worry that left only here they will fall through the cracks. I suggest we do either of these:

1. Do them now
2. Move them to issues
3. Write TODO comments in code

The order reflects my preference. You still have the context in your head and getting it done now seems the cheapest. This issue has already waited for a long time, and the number of impacted users I believe was low.

jdhoffa:
Happy to do them now, except unfortunately I am in calls all afternoon, and tomorrow is a holiday, so I will have to get to it on Friday.

R/summarize_weighted_production.R
```diff
@@ -53,37 +53,97 @@
 #'
 #' summarize_weighted_percent_change(master, use_credit_limit = TRUE)
 summarize_weighted_production <- function(data, ..., use_credit_limit = FALSE) {
+  summarize_weighted_production_(data, ..., use_credit_limit = use_credit_limit, add_targets = FALSE)
+}
+
+summarize_weighted_production_ <- function(data, ..., use_credit_limit = FALSE, add_targets = FALSE) {
```
jdhoffa, line 59: I have added this new internal function `summarize_weighted_production_` which contains the argument `add_targets`. I then call `summarize_weighted_production` with default values for that internal function (this is to preserve the existing arguments of `summarize_weighted_production`). 

@maurolepore let me know what you think of this solution as a way to avoid WET code

R/summarize_weighted_production.R
```diff
@@ -53,37 +53,97 @@
...
+    data %>%
+      summarize(
+        weighted_production = .data$production,
+        weighted_production_target = .data$production_target,
+        .groups = "keep"
+      ) %>%
+      ungroup(.data$technology) %>%
+      mutate(
+        weighted_technology_share = .data$weighted_production / sum(.data$weighted_production),
+        weighted_technology_share_target = .data$weighted_production_target / sum(.data$weighted_production_target)
```
maurolepore, line 137: The #FIXME comment is now misplaced. Maybe we can remove it and instead write a comment explaining why we use the "weighted_" prefix in this function? Or change the prefix for a temporary one?

maurolepore (approved):
LGTM. Thanks Jackson.

jdhoffa:
Exploring the failing CI checks right now. It seems they are on older R releases only?

R/target_market_share.R
```diff
@@ -300,7 +300,7 @@ target_market_share <- function(data,
   ald_with_benchmark <- calculate_ald_benchmark(ald, region_isos, by_company)
 
   data %>%
-    rbind(ald_with_benchmark) %>%
+    dplyr::bind_rows(ald_with_benchmark) %>%
```
jdhoffa, line 303: well hot diggity!

PR #305, opened 2021-06-07
Test that `target_market_share` fails with differing input currencies

### LINKED ISSUE #279: `target_market_share` should fail if input `loan_size_*_currency` isn't consistent
Currently, `target_market_share` requires `loan_size_*_currency` on input, and we group by it in several cases, however we have no possibility to actually convert currencies internally (so if the user inputs multiple currencies we would not be able to calculate targets). 

Instead, we should just check initially that all currencies are identical, and pass an informative error if not.

jdhoffa:
Closes #279

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # r2dii.analysis (development version)
 
+* `target_market_share()` now fails if input data has inconsistent values in the `loan_size_*_currency` (#279)
+
 # r2dii.analysis 0.1.8
 
 * `target_market_share()` now correctly outputs target `technology share`, in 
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # r2dii.analysis (development version)
 
+* `target_market_share()` now fails if input data has inconsistent values in the `loan_size_*_currency` (#279)
+
```
maurolepore, line 4: The error was already implemented, right? If all we did is to add tests, then there is no user-facing change to report in NEWS.

maurolepore (approved):
LGTM but this seems like an internal change with no user-facing consequences to report in NEWS.

PR #314, opened 2021-06-17
`target_sda` now outputs simplifed column names `metric` and `emission_factor`

### LINKED ISSUE #313: Simplify names output by `target_sda`
In particular, the most important columns output by `target_sda` are `emission_factor_metric` and `emission_factor_value`, whereas `target_market_share` outputs the more simple columns `metric` and `production` or `technology_share`. 

I propose:
* `emission_factor_metric` -> `metric`
* `emission_factor_value` -> `emission_factor`

``` r
library(dplyr)
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

matched <- loanbook_demo %>% 
  match_name(ald_demo) %>% 
  prioritize()

target_market_share(
  matched, 
  ald_demo, 
  scenario_demo_2020,
  region_isos_demo
) %>% 
  names()
#> [1] "sector"           "technology"       "year"             "region"          
#> [5] "scenario_source"  "metric"           "production"       "technology_share"

target_sda(
  matched, 
  ald_demo, 
  co2_intensity_scenario_demo
) %>% 
  names()
#> [1] "sector"                 "year"                   "emission_factor_metric"
#> [4] "emission_factor_value"
```

<sup>Created on 2021-06-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jdhoffa:
Closes #313 

Relates to 2DegreesInvesting/r2dii.plot#288
@MonikaFu


maurolepore (commented):
From a technical point of view this looks good but this is a breaking change that I think needs the OK from @AlexAxthelm. I also engaged @tposey28 to pick his brain. See https://github.com/2DegreesInvesting/r2dii.analysis/issues/313#issuecomment-863166681.

I suggest capturing everyone's thoughts in a public space -- maybe https://github.com/2DegreesInvesting/r2dii.data/discussions -- and take the change to try come up with a process that we can reuse elsewhere.

I would not merge until we hear at least from @AlexAxthelm

maurolepore:
@jdhoffa from conversations elsewhere I understand this PR is not crucial. Given it would impact r2dii.plot I would ask you to please hold until r2dii.plot is on CRAN. Anything that we do with this columns is internal and should be easy to adapt, but just now the change would complicate r2dii.plot, as we would have to add code to try handle the case with the old and new names.

jdhoffa:
@maurolepore totally happy to hold off. My thought was that it would be easier to make this change now before `r2dii.plot` is published, so that you would only have to handle this new case, and wouldn't have to change it later, but I see now that that, for `r2dii.plot` to remain backwards compatible, you would need to keep both cases!

PR #330, opened 2021-07-07
Group by region when calculating `technology_share`

### LINKED ISSUE #315: `coalcap` seems to have increasing `technology_share` after the `year` 2030, while the scenario still decreases
This is just a weird output using the `demo` datasets from `r2dii.data`. I am not sure what exactly is causing this, or what is the way forward, but it is a very peculiar result: 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

ald <- ald_demo %>% 
  filter(
    year > 2030,
    sector == "power"
  )

scenario <- scenario_demo_2020 %>% 
  filter(
    scenario == "sds",
    sector == "power",
    region == "global",
    year > 2030
  )

out <- loanbook_demo %>% 
  match_name(ald) %>% 
  prioritize() %>% 
  target_market_share(
    ald,
    scenario,
    region_isos_demo
  ) 

out %>% 
  filter(metric == "target_sds") %>% 
  select(technology, year, metric, technology_share)
#> # A tibble: 50 x 4
#>    technology  year metric     technology_share
#>    <chr>      <int> <chr>                 <dbl>
#>  1 coalcap     2031 target_sds            0.117
#>  2 coalcap     2032 target_sds            0.123
#>  3 coalcap     2033 target_sds            0.128
#>  4 coalcap     2034 target_sds            0.138
#>  5 coalcap     2035 target_sds            0.139
#>  6 coalcap     2036 target_sds            0.140
#>  7 coalcap     2037 target_sds            0.142
#>  8 coalcap     2038 target_sds            0.142
#>  9 coalcap     2039 target_sds            0.141
#> 10 coalcap     2040 target_sds            0.140
#> # … with 40 more rows
```

<sup>Created on 2021-06-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jdhoffa:
Closes #315 
Thanks @Antoine-Lalechere


R/summarize_weighted_production.R
```diff
@@ -262,14 +262,26 @@ add_percent_change <- function(data) {
 
 add_technology_share <- function(data) {
   data %>%
-    group_by(.data$sector_ald, .data$year, .data$scenario, .data$name_ald) %>%
+    group_by(
```
AlexAxthelm, line 265: Non-blocking: prefer  `[` to `$` as  an accessor (`.data["sector_ald"]`). 

Also, am  I missing a new best practice? why are we expressing this with `.data`?

AlexAxthelm (approved):

R/summarize_weighted_production.R
jdhoffa: Extracting "Non-blocking: prefer [ to $ as an accessor (.data["sector_ald"])." as a new issue as well, since this would require a rather large style change in the whole package (to keep consistency). 

Regarding `.data`, we would need @maurolepore to chip in, as I am actually not entirely sure. My understanding is that, when using piping, you should explicitly namespace

jdhoffa: See #335

maurolepore: In packages the `.data` pronoun avoids a R CMD NOTE about undefined global variables. You can also define each column you use with `globalVariables(c("column1", "column2"))` or initialize `column1 <- NULL; column2 <- NULL` before the first use, but the `.data` pronoun is the way  the tidyverse recommends. There is some article somewhere -- if curious let me know and I can try find it.

PR #337, opened 2021-07-08
`target_market_share` no longer allows for negative `production` targets

### LINKED ISSUE #336: `target_market_share` outputs negative values of `production`
Currently, there are cases in which `target_market_share` outputs negative values in the `production` column (see reprex).  
I would expect there to be no negative values in this column. 

This seems to happen when:
- the input `matched` and `ald` contain companies with only `brown` technologies AND
- the scenario decreases for a `green` technology

Since #328 `target_market_share` will now add green technologies, even if none were present initially. Since the target for green technologies is calculated using this equation: 
Screenshot 2021-07-08 at 13 36 23

we can see that if initially there was no production (p_i(t0) = 0) and the scenario decreases (s_i(t) > s_i(t0)), that we will be subtracting from 0, thus resulting in negative values. 

This is a methodological issue, which requires an interim technical solution. 

---

``` r
suppressPackageStartupMessages(library(dplyr))
library(r2dii.data)
library(r2dii.analysis)

matched <- tibble::tribble(
)

ald <- tibble::tribble(
  "shaanxi auto", "automotive",       "ice",  2025,           1,                1,            "BF",               TRUE,
  "shaanxi auto", "automotive",       "ice",  2026,           1,                1,            "BF",               TRUE
)

scenario <- tibble::tribble(
  "sds", "automotive",       "ice", "global",  2025,   0.5,     0,      "demo_2020",
  "sds", "automotive",  "electric", "global",  2025,   0.5,     0,      "demo_2020",
  "sds", "automotive",       "ice", "global",  2026,   0.5,    -1,      "demo_2020",
  "sds", "automotive",  "electric", "global",  2026,   0.5,    -1,      "demo_2020"
)

out <- target_market_share(
  matched,
  ald,
  scenario,
  region_isos_demo
)

out %>% 
  filter(metric == "target_sds") %>% 
  select(technology, year, production)
#> # A tibble: 4 x 3
#>   technology  year production
#>   <chr>      <dbl>      <dbl>
#> 1 electric    2025        0  
#> 2 electric    2026       -1  
#> 3 ice         2025        0.5
#> 4 ice         2026        0.5
```

<sup>Created on 2021-07-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jdhoffa:
Any `smsp` targets that are calculated negatively, will be attenuated to 0. 

This is a **technical** solution but does not solve the larger **methodological** issue, which is to be discussed by the analyst team in the next PACTA bi-weekly. 

Prior to merging this, I would like approval from @georgeharris2deg that he has read and understood this technical interim solution. 

Closes #336

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # r2dii.analysis (development version)
 
+* `target_market_share()` now never calculates `smsp` targets that are less than 
+  zero (#336). 
+
 * `target_market_share()` now only outputs `sector` values that are present in 
   all of `data`, `ald` and `scenario` (#329). 
 
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -202,6 +202,13 @@ target_market_share <- function(data,
       smsp_target_production = .data$initial_technology_production +
         (.data$initial_sector_production * .data$smsp)
     ) %>%
+    mutate(
+      smsp_target_production = ifelse(
+        smsp_target_production < 0,
+        0,
+        smsp_target_production
+      )
+    ) %>%
     select(
       -c(
         .data$tmsr,
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -204,9 +204,9 @@ target_market_share <- function(data,
     ) %>%
     mutate(
       smsp_target_production = ifelse(
-        smsp_target_production < 0,
+        .data$smsp_target_production < 0,
         0,
-        smsp_target_production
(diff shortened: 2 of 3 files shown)


georgeharris2deg:
Hi Jackson - Thanks for checking and yes I agree that we can proceed with this solution
Thanks 
@jdhoffa

AlexAxthelm (approved):

Issue #306, opened 2021-06-08
`target_market_share` passes an unhelpful internal error when a company has 0 production and `NA` emission_factors


jdhoffa:
In `target_market_share`, there is a downstream function `add_technology_share` which calculates the share of each technology, per company, year and sector. If, for some year, sector, company, the total production is 0, then the `add_technology_share` function will divide by 0, causing an error. 

We should likely add a check higher in the callstack to ensure we account for this situation. 

Thanks @EvgenyPetrovsky for flagging this, and drafting the useful reprex below: 

---

Brief description of the problem

``` r
library(magrittr)
require(r2dii.data)
#> Loading required package: r2dii.data
require(r2dii.match)
#> Loading required package: r2dii.match
require(r2dii.analysis)
#> Loading required package: r2dii.analysis

lbk_demo <- r2dii.data::loanbook_demo
ald_demo <- r2dii.data::ald_demo
scn_demo <- r2dii.data::scenario_demo_2020
rgn_demo <- r2dii.data::region_isos_demo

# matching
lbk_matched <- r2dii.match::match_name(loanbook = lbk_demo, ald = ald_demo) %>% 
  r2dii.match::prioritize()

# selecting first name from matched ald data and taking some year
one_name <- lbk_matched$name_ald %>% 
  unique() %>% 
  (function(x) x[!is.na(x)]) %>% 
  head(1)

one_year <- 2020

# breaking data across all technologies for one selected ald name_company and year
ald_broken <- ald_demo

ald_broken[ald_broken$name_company == one_name & ald_broken$year == one_year, "emission_factor"] <- NA

ald_broken[ald_broken$name_company == one_name & ald_broken$year == one_year, "production"] <- 0

# running analysis
target_market_share <- r2dii.analysis::target_market_share(
  
  data = lbk_matched, 
  
  ald = ald_broken, 
  
  scenario = scn_demo, 
  
  region_isos = rgn_demo,
  
  by_company = FALSE)
#> Error: Column `technology_share` must not contain any `NA`s.
```

<sup>Created on 2021-06-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

jdhoffa:
I think the easiest solution here is to simply filter the input `ald` file of all rows with "0" production. These entries are not used in the `target_market_share` calculation in any case.

jdhoffa:
# Summary
[...]

PR #348, opened 2021-07-26
For companies with 0 production in all sectors, set `technology_share` to 0 to avoid `NA`s 

Antoine-Lalechere:


R/summarize_weighted_production.R
```diff
@@ -269,7 +268,9 @@ add_technology_share <- function(data) {
       .data$name_ald,
       .data$region
     ) %>%
-    mutate(technology_share = .data$production / sum(.data$production)) %>%
+    mutate(prod_total_company=sum(.data$production),
+           tech_share=.data$production / sum(.data$production),
+           technology_share = ifelse(prod_total_company!=0,tech_share,0)) %>%
```
jdhoffa, line 273: NIT: Since we don't use `prod_total_company` again in the pipeline, it's better to set it to a temporary variable (prefixed by a `.`), and then set that variable back to `NULL` at the end of `mutate()`, so we don't pollute the dataset with unnecessary columns: 
``` r 
mutate(
   .x = sum(.data$production),
   technology_share = ifelse(
      .x == 0,
      0,
      .data$production / .x
   ),
   .x = NULL
)
```

R/summarize_weighted_production.R
```diff
@@ -282,7 +283,9 @@ add_technology_share_target <- function(data) {
       .data$name_ald,
       .data$region
     ) %>%
-    mutate(technology_share_target = .data$production_target / sum(.data$production_target)) %>%
+    mutate(prod_total_company_target=sum(.data$production_target),
+           tech_share_target = .data$production_target / sum(.data$production_target),
+           technology_share_target = ifelse(prod_total_company_target!=0, tech_share_target, 0)) %>%
```
jdhoffa, line 288: See above

tests/testthat/test-target_market_share.R
```diff
@@ -1143,3 +1143,24 @@ test_that("`target_market_share` outputs only positive values of `production`(#3
 
   expect_false(any(out$production < 0))
 })
+
+test_that("NA in technology share (ADO876)", {
+  ald <- tibble::tribble(
+    ~company_id, ~name_company, ~is_ultimate_owner,  ~sector,           ~technology, ~year, ~plant_location, ~production,   ~production_unit,  ~emission_factor,               ~emission_factor_unit, ~ald_timestamp,
+    38998,  "parc eolien de bel air",              FALSE, "power", "renewablescap",  2020,            "FR",      0, "MW",  0.74613363337134, "tonnes of CO2 per MWh",      "2021 Q1",
+    38998,  "parc eolien de bel air",              FALSE, "power", "renewablescap",  2021,            "FR",      1, "MW",  0.74613363337134, "tonnes of CO2 per MWh",      "2021 Q1"
+  )
```
jdhoffa, line 1152: this can be replace with the convenient helper function: 
``` r 
ald <- fake_ald(year = c(2020, 2021), production = c(0,1))

scenario <- fake_scenario(year = c(2020, 2021))

out <- target_market_share(fake_matched(), ald, scenario, region_isos_stable)

...
```

tests/testthat/test-target_market_share.R
```diff
@@ -1143,3 +1143,24 @@ test_that("`target_market_share` outputs only positive values of `production`(#3
...
+    ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,            ~level, ~score,      ~sector,      ~name_ald,  ~sector_ald,
+    "L162",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "power", "parc eolien de bel air", "power"
+  )
+  out <- target_market_share(
+    matched,
+    ald,
+    scenario_demo_2020,
+    region_isos_demo)
+
+  expect_equal(any(out$technology_share),"NA")
```
jdhoffa, line 1163: This isn't testing for what you think it is testing for. 
What this test is saying is: 
`any(out$technology_share)` asks if any of the values of `technology_share` are `TRUE` (which doesn't make sense, since `technology_share` is numeric.). So this condition will always output `FALSE`

Then `expect_equal(FALSE, "NA")` will always fail, since `FALSE != "NA"`. 

Try to think of a more specific test. This is always the MOST crucial part of the debugging process (IMO), so it's important to think carefully about this at the beginning.

The process should be:

* add failing test
* debug
* submit PR once this test passes

If you look at the CHECKS on this PR, you will see that they all fail, since this PR is failing.

jdhoffa (changes requested):
See comments. We can chat about the review today at 2pm

tests/testthat/test-target_market_share.R
jdhoffa: Before approving, this test will need to be something that: 
* fails on `master` version of `r2dii.analysis`
* passes on this branches version of `r2dii.analysis`

jdhoffa:
Closes #306

jdhoffa:
This PR seems to have a strange commit history, that's causing issues. 
Closing in favour of #350

PR #350, opened 2021-07-28
Set `technology_share` to 0 for companies with zero initial sectoral production

### LINKED ISSUE #306: `target_market_share` passes an unhelpful internal error when a company has 0 production and `NA` emission_factors

jdhoffa:
This PR is a sparse implementation of @Antoine-Lalechere's fix to set `technology_share` values to `0` should they have no initial sectoral production (#348). 

Closes #306 
Closes [ADO 876](https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/876)

### DIFF
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -269,7 +269,11 @@ add_technology_share <- function(data) {
       .data$name_ald,
       .data$region
     ) %>%
-    mutate(technology_share = .data$production / sum(.data$production)) %>%
+    mutate(
+      .x = sum(.data$production),
+      technology_share = ifelse(.data$.x == 0, 0, .data$production / .data$.x),
+      .x = NULL
+      ) %>%
     group_by(!!!dplyr::groups(data))
 }
 
@@ -282,7 +286,11 @@ add_technology_share_target <- function(data) {
       .data$name_ald,
       .data$region
     ) %>%
-    mutate(technology_share_target = .data$production_target / sum(.data$production_target)) %>%
+    mutate(
+      .x = sum(.data$production_target),
+      technology_share_target = ifelse(.data$.x == 0, 0, .data$production_target / .data$.x),
+      .x = NULL
+      ) %>%
     group_by(!!!dplyr::groups(data))
 }
 
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -273,7 +273,7 @@ add_technology_share <- function(data) {
       .x = sum(.data$production),
       technology_share = ifelse(.data$.x == 0, 0, .data$production / .data$.x),
       .x = NULL
-      ) %>%
+    ) %>%
     group_by(!!!dplyr::groups(data))
 }
 
@@ -290,7 +290,7 @@ add_technology_share_target <- function(data) {
       .x = sum(.data$production_target),
       technology_share_target = ifelse(.data$.x == 0, 0, .data$production_target / .data$.x),
       .x = NULL
-      ) %>%
+    ) %>%
     group_by(!!!dplyr::groups(data))
 }
 
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -402,8 +402,8 @@ calculate_ald_benchmark <- function(ald, region_isos, by_company) {
       .data$sector, .data$year, .data$region, .data$source
     ) %>%
     mutate(
-      .x = .data$production,
-      technology_share = .data$.x / sum(.data$.x),
+      .x = sum(.data$production),
+      technology_share = ifelse(.data$.x == 0, 0, .data$production / .data$.x),
       .x = NULL,
       metric = "corporate_economy"
     ) %>%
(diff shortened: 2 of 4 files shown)


R/summarize_weighted_production.R
```diff
@@ -282,7 +286,11 @@ add_technology_share_target <- function(data) {
       .data$name_ald,
       .data$region
     ) %>%
-    mutate(technology_share_target = .data$production_target / sum(.data$production_target)) %>%
+    mutate(
+      .x = sum(.data$production_target),
+      technology_share_target = ifelse(.data$.x == 0, 0, .data$production_target / .data$.x),
+      .x = NULL
+    ) %>%
```
maurolepore, line 293: Just an idea, I wonder if we could extract this pattern into a wrapper around `mutate()` with a new `.x` argument that will be automatically destroyed  after use. Nothing to take action but to spark thoughts.

R/target_market_share.R
```diff
@@ -402,8 +402,8 @@ calculate_ald_benchmark <- function(ald, region_isos, by_company) {
       .data$sector, .data$year, .data$region, .data$source
     ) %>%
     mutate(
-      .x = .data$production,
-      technology_share = .data$.x / sum(.data$.x),
+      .x = sum(.data$production),
```
maurolepore, line 405: Every time I see sum I think of `na.rm`. Maybe `NA`s are handled above but one way to remove the worry is to either add `na.rm = TRUE` explicitely (if it corresponds) or maybe implement `add()` as a wrapper of `sum()` that defaults to `na.rm = TRUE`.

R/target_market_share.R
```diff
@@ -437,11 +437,23 @@ reweight_technology_share <- function(data, ...) {
...
+      .sum_x = sum(.data$.x),
       .y = .data$weighted_technology_share_target,
-      weighted_technology_share = .data$.x / sum(.data$.x),
-      weighted_technology_share_target = .data$.y / sum(.data$.y),
+      .sum_y = sum(.data$.y),
+      weighted_technology_share = ifelse(
+        .data$.sum_x == 0,
+        0,
+        .data$.x / .data$.sum_x
+        ),
```
maurolepore, line 447: Nit: Is it worth taking so much vertical screen-real-state?

tests/testthat/test-target_market_share.R
```diff
@@ -1143,3 +1143,34 @@ test_that("`target_market_share` outputs only positive values of `production`(#3
 
   expect_false(any(out$production < 0))
 })
+
+test_that("`target_market_share` outputs as expected for companies with 0 initial sectoral production (#306)", {
```
maurolepore, line 1147: Nit: Do we need to wrap at 80 chars?

maurolepore (approved):
LGTM. The tests make me comfortable that the code does what's intended to do. My comments are all nit.

R/summarize_weighted_production.R
jdhoffa: Ya this seems like an interesting idea. Would have to think carefully about what to call the new function

R/target_market_share.R
jdhoffa: Nope! You are right, this was silly.

R/target_market_share.R
jdhoffa: I am a little hesitant to pass `na.rm = TRUE` to calls to `sum()`, my reasoning is as follows: 
#306 was flagged because `target_market_share` passing an error, due to some `NA` that was added internally. The input data had no `NA`s, so it was something in the code that was introducing the `NA`s. 

These `NA`s were not expected, so this error forced us to recognize the problem and address it. 

If we had passed `na.rm = TRUE`, I imagine this bug would have persisted silently, and we likely would have taken a lot longer to learn about it.

tests/testthat/test-target_market_share.R
jdhoffa: Done

PR #357, opened 2021-07-30
Use testthat edition 3

### LINKED ISSUE #354: Use testthat edition 3

maurolepore:
Closes #354

﻿- Use thestthat 3e
- When testing, make library() quiet
- Maybe FIXME tolerance
- Avoid message while testing
- Use expect_is_s3_class()
- Use all_of()
- Use expect_snapshot()


tests/testthat/_snaps/target_market_share.md
```diff
@@ -0,0 +1,5 @@
+# with fake data outputs known value
+
+    Code
+      # out
+
```
jdhoffa, line 5: Does this snapshot have no output? That's interesting

tests/testthat/test-target_market_share.R
```diff
@@ -49,15 +49,14 @@ test_that("with fake data outputs known value", {
     fake_scenario(),
     region_isos_stable
   )
-
-  expect_known_value(out, "ref-target_market_share", update = FALSE)
+  expect_snapshot("out")
```
jdhoffa, line 52: This `out` should be an object, not a character. This is what is causing the empty output in tests/testthat/_snaps/target_market_share.md

tests/testthat/test-target_market_share.R
```diff
@@ -457,10 +458,11 @@ test_that("with known input outputs `technology_share` as expected (#184, #262)"
     c(0.6875, 0.3125)
   )
 
+  FIXME <- 1e-2  # Do we need 1e-3?
   expect_equal(
     out$corporate_economy$technology_share,
     c(0.666, 0.333),
-    tolerance = 1e-3
+    tolerance = FIXME
```
jdhoffa, line 465: What is the rationale for making the tolerance lower? 
The expected values are `c(0.666666666666..., 0.33333333333333...)`. My preference would be to have a tolerance that includes more significant digits, to ensure we don't get a false pass (but the choice of `1e-3` was entirely arbitrary).

jdhoffa: This applies to the next test as well

jdhoffa (changes requested):

tests/testthat/test-target_market_share.R
maurolepore: Thanks!

tests/testthat/test-target_market_share.R
maurolepore: I made it lower only because otherwise the test failed. And nothing changed in the code so I thought it's something to hack for now.

With tolerance `1e-3` I get:

```r
  `actual`: 0.6667 0.3333
`expected`: 0.6660 0.3330

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 52 ]
```

jdhoffa: Ah interesting, I think I misunderstood the meaning of `tolerance` in that case. To me `1e-2` would mean two significant digits (or `0.66`), and `1e-3` would mean `0.666`, but I guess it's adding 1 extra digit tolerance. 

In that case, `1e-2` is totally fine.

jdhoffa (approved):
LGTM

PR #358, opened 2021-07-30
MERGE 2nd: Make sysdata reproducible

### LINKED ISSUE #353: Make sysdata reproducible
If I source data-raw/sysdata.R I get an error:

```r
> loanbook_stable <- left_join(
+   loanbook_demo, nace_classification,
+   by = c(sector_classification_direct_loantaker = "code")
+ )
Error: Can't join on `x$sector_classification_direct_loantaker` x `y$sector_classification_direct_loantaker` because of incompatible types.
ℹ `x$sector_classification_direct_loantaker` is of type <double>>.
ℹ `y$sector_classification_direct_loantaker` is of type <character>>.
```

The message suggests the solution is to coerse loanbook_demo$sector_classification_direct_loantaker
to a character.

This change will impact regression tests.﻿

maurolepore:
This PR merges not into master but into 354_*, which is
a way to say it depends on #354.

Closes #353

When I tried to source data-raw/sysdata.R I got an error,
showing that sysdata was no longer reproducible. This pr
makes sysdata reproducible again, by ensuring two columns
we want to join by of the same type.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,6 @@ Imports:
     rlang,
     tidyr,
     tidyselect,
-    withr,
     zoo
 Suggests: 
     covr,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,6 +42,7 @@ Imports:
     rlang,
     tidyr,
     tidyselect,
+    withr,
     zoo
 Suggests: 
     covr,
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -42,6 +42,7 @@ Imports:
     rlang,
     tidyr,
     tidyselect,
+    withr,
```
jdhoffa, line 45: Do we need to depend on a package that is used only in a sourcing file in `data-raw`?

maurolepore: No, thanks!!

jdhoffa (approved):
LGTM

PR #359, opened 2021-07-30
MERGE 1st. Avoid internal use of r2dii.match

### LINKED ISSUE #356: Use r2dii.match as minimally as possible

maurolepore:
Closes #356

This depends on #358 so it merges into it.

Here avoid using r2dii.match in tests. We still use it in
examples, but we might not do so by instead using the
(now internal) dataset matched_stable.﻿


jdhoffa:
Not a full review, but my immediate thoughts are regarding the `*_stable` datasets. 
It seems to me, rather than go through the effort of "freezing" `r2dii.data v0.1.4`, it might make more sense to just add more `fake_*()` helper functions? 

The only benefit I can think of using `r2dii.data` datasets, is that they would automatically update and let us now if we had issues with `r2dii.data` and `r2dii.analysis` working together (integration tests). If we are going to freeze `r2dii.data`, and forego that benefit, then I don't see a downside to just defining them explicitly in `r2dii.analysis`, does that make sense?

maurolepore:
> Regarding the *_stable datasets ... it might make more sense to just add more `fake_*()` helper functions?

I think the `stable_*` and `fake_*()` datasets have different goals:

* `fake_*()` help create tiny datasets, to test specific properties of the functions in the package.

* `stable_*` provide large datasets for snapshot tests, to catch unspecific issues we don't know about until we find a test that fails.

While both seem to have a purpose, I think the regression tests have been too brittle, so I would consider dropping them. With that we wouldn't need the `stable_*` datasets.

jdhoffa:
Yes, I agree completely with what you wrote above. The `stable_*` datasets, and the tests that use them don't seem to add a lot to the package. 

My thoughts are as follows: 
* It is useful to have unit tests using tiny datasets that test specific functions in `r2dii.analysis` (so it is useful to keep `fake_*` type datasets
* It is useful to ensure the `r2dii.analysis` integrates with the LATEST `r2dii.data` and `r2dii.match` outputs (integration tests). For this, the `stable_*` don't really achieve this purpose since they would only represent a snapshot of what the datasets look like
* It is useful to test the ACTUAL data that Asset Resolution sets out (this data is modeled on the `r2dii.data` data, but still would be good to have some integration testing framework using the real banks release datasets. These are private, so this would need to be some kind of internal testing infrastructure)

Are integration tests something that are often done using `testthat`? Or is it better to do this using some other infrastructure.

jdhoffa:
Closing in favor of #360

PR #367, opened 2021-08-16
Address CRAN submission notes

### LINKED ISSUE #366: CRAN Notes from latest re-submission:
Please reduce the length of the title to less than 65 characters.

Please add \value to .Rd files regarding exported methods and explain
the functions results in the documentation. Please write about the
structure of the output (class) and also what the output means. (If a
function does not return a value, please document that too, e.g.
\value{No return value, called for side effects} or similar)
Missing Rd-tags:
      pipe.Rd: \value
      tidyeval.Rd: \value

Additionally:
Have the issues why your package was archived been fixed?
Please explain this in the submission comments.

Please fix and resubmit.

jdhoffa:
Closes #366

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,5 @@
 Package: r2dii.analysis
-Title: Tools to Measure Climate Scenario Alignment of Corporate Loans
-    Portfolios
+Title: Measure Climate Scenario Alignment of Corporate Loans
 Version: 0.1.10.9000
 Authors@R: 
     c(person(given = "Jackson",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: r2dii.analysis
-Title: Tools to Measure Climate Scenario Alignment of Corporate Lending
+Title: Tools to Measure Climate Scenario Alignment of Corporate Loans
     Portfolios
 Version: 0.1.10.9000
 Authors@R: 
--- a/man/r2dii.analysis-package.Rd
+++ b/man/r2dii.analysis-package.Rd
@@ -4,8 +4,7 @@
 \name{r2dii.analysis-package}
 \alias{r2dii.analysis}
 \alias{r2dii.analysis-package}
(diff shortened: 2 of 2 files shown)


jdhoffa:
This comment is interesting: 

> Please add \value to .Rd files regarding exported methods and explain
the functions results in the documentation. Please write about the
structure of the output (class) and also what the output means. (If a
function does not return a value, please document that too, e.g.
\value{No return value, called for side effects} or similar)
Missing Rd-tags:
pipe.Rd: \value
tidyeval.Rd: \value

It seems like the `utils` functions we add need to have a `\value` entry? Odd that it didn't cause problems before.

jdhoffa:
It's also a bit surprising, since I looked at `dplyr` and `ggplot2` and they both use `utils-tidyeval` without a `\value` in their docs.

DESCRIPTION
```diff
@@ -1,5 +1,5 @@
 Package: r2dii.analysis
-Title: Tools to Measure Climate Scenario Alignment of Corporate Lending
+Title: Tools to Measure Climate Scenario Alignment of Corporate Loans
```
maurolepore, line 2: Nit: I think it's best to remove "Tools to" as there is a record that CRAN considered that uneccessary (as every package provides tools to do something).

https://github.com/DavisVaughan/extrachecks under _Your package DESCRIPTION Title is flagged._:

> I’ve been flagged for a “redundant” title. I had: “A Toolkit for the Construction of Modeling Packages” which was flagged since “Toolkit for” seemed redundant. I changed it to “Construct Modeling Packages” and was accepted.

maurolepore (approved):
It seems that CRAN has become a bit more rigurous recently, and so far they seem to apply the increased rigour to new packages -- not existing ones.

> It seems like the utils functions we add need to have a \value entry? Odd that it didn't cause problems before.

> It's also a bit surprising, since I looked at dplyr and ggplot2 and they both use utils-tidyeval without a \value in their docs.

Some of those new requirements are captured at https://github.com/DavisVaughan/extrachecks

PR #368, opened 2021-08-16
Add return values for pipe and tidyeval

maurolepore:
Complements #367 

I updated usethis and rerun `use_pipe()` and `use_tidyevel()` hoping the latest version of usethis would add `@return` to those helpfiles. That was the case for pipe but not for tidyeval. 

- Document return value of the pipe via the development version of `usethis::use_pipe()` 
- Run `usethis::use_tidyeval()` plus move things around a bit and add `@return` to meet CRAN's requirement.

I'm still unsure what's the recommended way to deal with this. I searched on rOpenSci's slack and found nothing useful. We may come across some other way but for now this should work.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,7 +39,7 @@ Imports:
     glue,
     magrittr,
     r2dii.data,
-    rlang,
+    rlang (>= 0.1.2),
     tidyr,
     tidyselect,
     zoo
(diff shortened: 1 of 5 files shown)


maurolepore:
I'm merging with no review because:
* This is a hotfix (try bring this package back to CRAN) 
* It touches no production code.
* Jackson expect this change.

jdhoffa:
lgtm

PR #374, opened 2021-09-21
Simpler News

maurolepore:
The News menu has a "Releases" section that I now find confusing
because there are more releases than entries under News/Releases,
pointing to external articles. But we don't write articles for
most small patches.

This PR removes the headings "Releases" and "Changelog", to
instead expose the changelog directly under the News navbar.

The goal is to reduce confusion and to reduce maintenance cost,
by not feeling we need to add an article for small patches. 

An alternative could be to add a single link to the blog, so
users can discover it and then serve themselves what they 
need. 

--

Relates to similar PRs in other r2dii packages (e.g. https://github.com/2DegreesInvesting/r2dii.match/pull/382) but we can discuss only here and then I can apply the decision to all repos.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -43,15 +43,7 @@ navbar:
       href: articles/production-percent-change.html
 
   - text: "News"
-    menu:
-    - text: "Releases"
-    - text: "Version 0.1.0"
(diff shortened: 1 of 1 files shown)


jdhoffa:
This looks good to me, and i back not maintaining things that we don't really use. 

The option of adding a single link to the blog could be a good idea, as we can then pick and choose which releases deserve an explanation.

jdhoffa (approved):

PR #379, opened 2021-11-08
Update tmsr explanation in article

jdhoffa:
In the current formulation, the TMSR and SMSP approaches seem very different. By expanding the equation and showing the derivation, I hope to make it more clear to users how similar they actually are. 

@maurolepore not sure if I need to build pkgdown or anything to incorporate these changes? I think this is all done on CI right?


maurolepore:
Yeah, I think CI workers will do it for you.

maurolepore (approved):
LGTM. Thanks!

PR #383, opened 2022-03-16
`target_market_share` gains columns for percentage of initial production by (sector or technology)

jdhoffa:
This PR adds two columns to the usual output of `target_market_share()`, namely: 
* `percentage_of_initial_production_by_scope`
* `scope`

(I am not particularly happy with those names, but couldn't think of a better one for now)

These columns are meant to calculate the "new" plotting methodology (ADO 4454), that calculates two different percent growths depending on if the Technology Market Share or Sector Market Share is applied. 

In the case of high-carbon technologies (or technologies for which the Technology Market Share is applied), the following percentage of initial production is calculated:
Screenshot 2022-03-16 at 15 31 19
which is basically the percent change in production, as a percentage of the initial production in that technology. 

In the case of low-carbon technologies (or technologies for which the Sector Market Share is applied), the percentage is take as a percentage of the initial sectoral production: 
Screenshot 2022-03-16 at 15 33 47

This PR is the first step in getting these values plotted, similar adjustments may need to be made in 2DegreesInvesting/r2dii.plot to allow these values to be properly plotted. 

FYI. @MonikaFu 

Relates to ADO 4143


maurolepore:
Here are some early thoughts -- I see this PR is still a draft.

These two columns seem related, based on the suffix:

* `percentage_of_initial_production_by_scope`
* `scope`

I think the relationship is more clearly seen in a prefix, rather than suffix. It seems that both columns relate to the concept of "production" and then to the existing column `production`. So my suggestion would be this:

* `production_<unit>` = `production`
* `production_initial_pcnt` = `percentage_of_initial_production_by_scope`
* `production_scope` = `scope`.

cjyetman (approved):
This is very hard to review from an output perspective without sufficient context, but generally I understand the logic and it seems sound. All comments I leave are nits which can be ignored.

README render action failed. That may be from a temporary error during the setup, which might simply require a re-run... or it may be due to using the v1 r-lib actions which should probably be updated (though I'd suggest doing that in a separate PR).

R/target_market_share.R
```diff
@@ -233,8 +233,7 @@ target_market_share <- function(data,
         green_or_brown = "green_or_brown"
       )
     ) %>%
-    warn_if_has_zero_rows("Joining `r2dii.data::green_or_brown` outputs 0 rows") %>%
-    select(-.data$target_name, -.data$green_or_brown)
+    warn_if_has_zero_rows("Joining `r2dii.data::green_or_brown` outputs 0 rows")
```
cjyetman, line 236: seems a bit strange to throw a warning here, but then act on the situation (return an empty data frame) later

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
   data <- data %>%
     pivot_wider2()
 
+  data <- data %>%
+    left_join(green_or_brown, by = c("sector", "technology")) %>%
+    left_join(tmsr_or_smsp, by = "green_or_brown") %>%
```
cjyetman, line 320: because these are objects defined above (not shown by default in this diff, it's hard to know if they should exist here... it would be easier to read/understand if you call the objects/functions directly here since they don't appear to be modified anywhere above (`r2dii.data::green_or_brown` and `tmsr_or_smsp()`)

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+    rename(target_name = .data$which_metric) %>%
+    select(-.data$green_or_brown)
+
+  data <- data %>%
+    dplyr::bind_rows(relevant_corporate_economy)
+
+  percent_by_sector_groups <- maybe_add_name_ald(
+    c("sector", "region", "scenario_source", "metric"),
+    by_company
+  )
```
cjyetman, line 343: very hard to guess what this `maybe_add_name_ald()` function does

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+
+  data <- data %>%
+    group_by(!!!rlang::syms(c(percent_by_sector_groups, "technology"))) %>%
+    arrange(.data$year) %>%
+    mutate(initial_technology_production = first(.data$production)) %>%
```
cjyetman, line 348: I generally prefer to use the `order_by` argument of `dplyr::first()` for this pattern to avoid changing anything unnecessary in the data frame

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+    group_by(!!!rlang::syms(c(percent_by_sector_groups, "year"))) %>%
+    mutate(
+      .sector_production = sum(.data$production)
+      ) %>%
```
cjyetman, line 352: why is this spread across 3 lines?

also consider if you need to use `na.rm = TRUE` in this `sum()`

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+    group_by(!!!rlang::syms(percent_by_sector_groups)) %>%
+    mutate(
+      initial_sector_production = first(.data$.sector_production),
```
cjyetman, line 355: whenever I use `dplyr::first()`, I prefer to be explicit about the ordering using the `order_by` argument

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+      .sector_production = NULL
```
cjyetman, line 356: generally I prefer to use the `select(-{varname})` syntax for removing variables (as you've done elsewhere in this PR) because I feel it is more obvious/explicit what it does

R/target_market_share.R
```diff
@@ -316,15 +315,72 @@ target_market_share <- function(data,
...
+      percentage_of_initial_technology_production = (.data$production - .data$initial_technology_production) / .data$initial_technology_production,
+      percentage_of_initial_sector_production  = (.data$production - .data$initial_technology_production) / .data$initial_sector_production,
+      initial_technology_production = NULL,
+      initial_sector_production = NULL
+    ) %>%
+    mutate(
+      scope = dplyr::case_when(
+        target_name == "tmsr_target_production" ~ "technology",
+        target_name == "smsp_target_production" ~ "sector"
+      ),
```
cjyetman, line 369: consider whether it's important to have a default value here

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+  out <- target_market_share(
+    fake_matched(),
+    fake_ald(),
+    fake_scenario(),
+    region_isos_stable
+  )
+
+  expected_added_columns <- c( "scope", "percentage_of_initial_production_by_scope")
+
+  expect_equal(setdiff(expected_added_columns, names(out)),character(0))
```
cjyetman, line 1191: is this not the same as `expect_equal(names(out), expected_added_columns)`?

(for me, it's a bit more logical to put the thing you're testing first, and the reference you're testing against 2nd

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+})
+
+test_that("with known input outputs `percent_of_initial_production_by_scope` as
```
cjyetman, line 1194: consider putting a comma after "input" to improve readability

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+          expected, at portfolio level", {
+            portfolio <- fake_matched(
+              name_ald = c("comp1", "comp2")
+            )
+
+            ald <- fake_ald(
+              technology = c("electric", "ice", "electric", "ice", "electric", "ice", "electric", "ice"),
+              year = c(2020, 2020, 2021, 2021, 2020, 2020, 2021, 2021),
```
cjyetman, line 1202: generally I believe it's preferred to use integers with the `2020L` syntax when relevant

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+              portfolio,
+              ald,
+              scenario,
+              region_isos_stable,
+              by_company = FALSE,
+              weight_production = FALSE
+            )
+
+            out_percent <- out %>%
+              filter(.$year == 2021) %>%
```
cjyetman, line 1224: you've used the `.data$` pronoun elsewhere in this PR... is there a reason for using the `.$` here?

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+    region_isos_stable
+  )
+
+  expected_added_columns <- c( "scope", "percentage_of_initial_production_by_scope")
+
+  expect_equal(setdiff(expected_added_columns, names(out)),character(0))
+})
+
+test_that("with known input outputs `percent_of_initial_production_by_scope` as
+          expected, at portfolio level", {
```
cjyetman, line 1195: something's weird with the indentation here?

tests/testthat/test-target_market_share.R
```diff
@@ -1175,3 +1177,135 @@ test_that("`target_market_share` outputs as expected for companies with 0
...
+              region_isos_stable,
+              by_company = FALSE,
+              weight_production = FALSE
+            )
+
+            out_percent <- out %>%
+              filter(.$year == 2021) %>%
+              split(.$metric)
+
+            expect_equal(
```
cjyetman, line 1227: `testthat::expect_equal` has a `tolerance` argument that could make things easier here (as opposed to rounding)

jdhoffa:
Thanks for the initial review guys. I realized that the logic of this is super hard to follow, and that this function needs a bit of a refactor to clarify what is going on, so I have submitted a pure refactor PR #384 

I will open this PR from draft mode once I clean up some of the tech debt that has accumulated.

R/target_market_share.R
jdhoffa: Ya this is a good point, going to think if there is a better way to do this so the actions are closer together.

R/target_market_share.R
jdhoffa: good point, I tried to clarify this in #384

R/target_market_share.R
jdhoffa: Very useful! thank you

R/target_market_share.R
jdhoffa: It is spread across 3 lines because I forgot to run the styler hehe >:)

Also I opted to have `target_market_share` be very picky, and fail if any crucial input columns contain `NA`s, so in principle, we shouldn't need to worry about that. If there are `NA`s, they will have been introduced by a bug in the code, and in general, I would rather know about that bug, than have it pass silently.

R/target_market_share.R
jdhoffa: Hmm, ya I was waffling between these two approaches actually. I do like the `.varname = blah, usefulvar = function(.varname), .varname = NULL`, when you are just creating an interim column that only gets used in one mutate call, but I think that only makes sense if it can all fit into the same mutate call. 

I think I agree that the `select()` approach works better here.

tests/testthat/test-target_market_share.R
jdhoffa: Not necessarily, the order of the columns could be different

tests/testthat/test-target_market_share.R
jdhoffa: Honestly I didn't even think about it hahaha

tests/testthat/test-target_market_share.R
cjyetman: it's more below.... looks like the entire function is indented multiple tabs?

cjyetman:
btw, what is the `ref-target_market_share` binary file with no extension?

jdhoffa:
> btw, what is the `ref-target_market_share` binary file with no extension?

Those are reference datasets generated automatically by `expect_known_value()`, used for regression tests. Basically, `expect_known_value(..., update = TRUE)` saves a binary file with whatever data you give to it, and then you can test that that file never changes. 

Honestly those tests are super brittle and not that useful (since often you actually do want the output to change slightly), I'll probably remove them all at some point.

R/target_market_share.R
jdhoffa: Ok I tried to be more explicit about these datasets. The thing is, they are used twice, and the functions are not terribly close together (they can't be really). So I'm not sure how to make it any more obvious where they come from, without importing them twice.

tests/testthat/test-target_market_share.R
jdhoffa: hmmm weird it looks ok in my RStudio?

jdhoffa:
Ok this is now ready for review. The main thing that is bothering me is the names of the final output columns `percentage_of_initial_prouduction_by_scope` and `scope`. They are accurate, but they feel clunky and confusing. If anyone has better ideas, please let me know.

cjyetman (approved):

MonikaFu (approved):
I didn't look super extensively into your implementation but LGTM. We've already discussed the column names and I don't have any better ideas, unfortunately. 

Are there new toy output data that I can use to update the toy data in `r2dii.plot`?

jdhoffa:
@MonikaFu I haven't included a toy output dataset in the package itself, but I think the easiest way to get an (extensive) one would be: 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

out <- loanbook_demo %>% 
  match_name(ald_demo) %>% 
  prioritize() %>% 
  target_market_share(ald_demo, scenario_demo_2020, region_isos_demo)

out %>% 
  select(percentage_of_initial_production_by_scope, scope)
#> # A tibble: 2,334 × 2
#>    percentage_of_initial_production_by_scope scope 
#>                                        <dbl> <chr> 
#>  1                                   0       sector
#>  2                                   0       sector
#>  3                                   0       sector
#>  4                                   0       sector
#>  5                                   0.00352 sector
#>  6                                   0.00108 sector
#>  7                                   0.00653 sector
#>  8                                   0.00137 sector
#>  9                                   0.00705 sector
#> 10                                   0.00213 sector
#> # … with 2,324 more rows
```

<sup>Created on 2022-03-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I imagine a change will need to be made to this dataset in `r2dii.plot` to reflect the new columns: https://github.com/2DegreesInvesting/r2dii.plot/blob/develop/data-raw/market_share.csv

MonikaFu:
@jdhoffa - sounds great, thanks

PR #384, opened 2022-03-17
Refactor `target_market_share()`

jdhoffa:
This PR is a very focused refactor of `target_market_share()` to tidy up some technical debt that has accumulated, and hopefully make it more clear how the function works. 

The diffs are large, however no tests were touched, and all tests + checks pass as expected, so we can be reasonably sure that the function still works as expected.

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -199,66 +199,12 @@ target_market_share <- function(data,
 
   target_groups <- c("sector_ald", "scenario", "region", "name_ald")
 
-  data <- data %>%
-    group_by(!!!rlang::syms(c(target_groups, "year"))) %>%
-    mutate(sector_production = sum(.data$production)) %>%
-    arrange(.data$year) %>%
-    group_by(!!!rlang::syms(target_groups)) %>%
-    mutate(initial_sector_production = first(.data$sector_production)) %>%
-    select(-.data$sector_production)
-
-  data <- data %>%
-    group_by(!!!rlang::syms(c(target_groups, "technology", "year"))) %>%
-    mutate(technology_production = sum(.data$production)) %>%
-    arrange(.data$year) %>%
-    group_by(!!!rlang::syms(c(target_groups, "technology"))) %>%
-    mutate(initial_technology_production = first(.data$technology_production)) %>%
-    select(-.data$technology_production) %>%
-    ungroup()
+  data <- calculate_targets(data, target_groups)
 
   green_or_brown <- r2dii.data::green_or_brown
   tmsr_or_smsp <- tmsr_or_smsp()
 
-  data <- data %>%
-    mutate(
-      tmsr_target_production = .data$initial_technology_production *
-        .data$tmsr,
-      smsp_target_production = .data$initial_technology_production +
-        (.data$initial_sector_production * .data$smsp)
-    ) %>%
-    mutate(
-      smsp_target_production = ifelse(
-        .data$smsp_target_production < 0,
-        0,
-        .data$smsp_target_production
-      )
-    ) %>%
-    select(
-      -c(
-        .data$tmsr,
-        .data$smsp,
-        .data$initial_technology_production,
-        .data$initial_sector_production
-      )
-    ) %>%
-    pivot_longer(
-      cols = c(
-        "tmsr_target_production", "smsp_target_production"
-      ),
-      names_to = "target_name",
-      values_to = "production_target"
-    ) %>%
-    left_join(tmsr_or_smsp, by = c(target_name = "which_metric")) %>%
-    inner_join(
-      green_or_brown,
-      by = c(
-        sector_ald = "sector",
-        technology = "technology",
-        green_or_brown = "green_or_brown"
-      )
-    ) %>%
-    warn_if_has_zero_rows("Joining `r2dii.data::green_or_brown` outputs 0 rows") %>%
-    select(-.data$target_name, -.data$green_or_brown)
+  data <- pick_sms_or_tms_target(data, green_or_brown, tmsr_or_smsp)
 
   if (nrow(data) == 0) {
     return(empty_target_market_share_output())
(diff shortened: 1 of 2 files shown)


R/target_market_share.R
```diff
@@ -282,105 +188,65 @@ target_market_share <- function(data,
...
-}
-
-tmsr_or_smsp <- function() {
-  dplyr::tribble(
-    ~which_metric, ~green_or_brown,
-    "tmsr_target_production", "brown",
-    "smsp_target_production", "green"
-  )
-}
+calculate_targets <- function(data){
```
cjyetman, line 207: nit: no space before "{"

R/target_market_share.R
```diff
@@ -282,105 +188,65 @@ target_market_share <- function(data,
...
+  target_groups <- c("sector_ald", "scenario", "region", "name_ald")
 
-separate_metric_from_name <- function(data) {
   data %>%
+    arrange(.data$year) %>%
+    group_by(!!!rlang::syms(c(target_groups, "technology", "year"))) %>%
+    mutate(technology_production = sum(.data$production)) %>%
```
cjyetman, line 213: nit: production data seems likely to have `NA`s in it, or would it always be 0?

R/target_market_share.R
```diff
@@ -282,105 +188,65 @@ target_market_share <- function(data,
...
+      initial_sector_production = NULL,
+      smsp = NULL
+    ) %>%
+    mutate(
+      smsp_target_production = ifelse(
+        .data$smsp_target_production < 0,
+        0,
+        .data$smsp_target_production
+      )
+    ) %>%
```
cjyetman, line 242: nit: I like using `pmax()` and/or `pmin()` for clamping vectors to a range, e.g.
```r
pmax(-1:3, 0)
#> [1] 0 0 1 2 3
```

R/target_market_share.R
```diff
@@ -282,105 +188,65 @@ target_market_share <- function(data,
...
+    pivot_longer(
+      cols = c(
+        "tmsr_target_production", "smsp_target_production"
+      ),
```
cjyetman, line 246: nit: is this longer than 80 characters?

R/target_market_share.R
```diff
@@ -419,6 +285,37 @@ calculate_ald_benchmark <- function(ald, region_isos, by_company) {
   out
 }
 
+add_name_ald_if_by_company <- function(data, by_company = FALSE) {
+  if (by_company) {
+    data <- c(data, "name_ald")
+  }
+
+  data
+}
```
cjyetman, line 294: nit: is `data` a vector here instead of a data frame as it seems to be most other places in this script? hard to tell without any validation

cjyetman: i.e. if it is a vector, I'd expect this to work properly `data <- c(data, "name_ald")`... but if it's a data frame, I'm not sure what would happen
```r
data <- data.frame(x = 1:2); c(data, "test")
#> $x
#> [1] 1 2
#> 
#> [[2]]
#> [1] "test"
```

R/target_market_share.R
```diff
@@ -470,6 +345,104 @@ aggregate_by_loan_id <- function(data) {
     ungroup()
 }
 
+format_output_dataframe <- function(data){
+  data <- data %>%
+    pivot_wider2(
```
cjyetman, line 350: didn't you delete this function above somewhere?

R/target_market_share.R
```diff
@@ -470,6 +345,104 @@ aggregate_by_loan_id <- function(data) {
...
+    separate_metric_from_name() %>%
+    pivot_wider2()
+}
+
+separate_metric_from_name <- function(data) {
+  data %>%
+    mutate(
+      name = sub("weighted_", "", .data$name),
+      name = sub("(production)_", "\\1-", .data$name),
+      name = sub("(technology_share)_", "\\1-", .data$name)
```
cjyetman, line 377: is "-" a proper "syntactic" character for data frame column name?

R/target_market_share.R
```diff
@@ -482,3 +455,57 @@ check_valid_columns <- function(data, valid_columns) {
 
   invisible(data)
 }
+
+pivot_wider2 <- function(data, ...) {
+  abort_if_has_list_colums(data)
+
+  out <- suppressWarnings(pivot_wider(data, ...))
```
cjyetman, line 462: out of curiosity, what warnings are you expecting?

cjyetman (approved):
made some non-critical comments, but LGTM

R/target_market_share.R
```diff
@@ -98,45 +98,9 @@ target_market_share <- function(data,
...
 
-  check_valid_columns(data, valid_columns)
-
-  crucial_scenario <- c("scenario", "tmsr", "smsp")
-  check_crucial_names(scenario, crucial_scenario)
-  check_crucial_names(ald, "is_ultimate_owner")
-  walk_(crucial_scenario, ~ check_no_value_is_missing(scenario, .x))
-
-  data <- aggregate_by_loan_id(data)
+  data <- aggregate_by_name_ald(data)
```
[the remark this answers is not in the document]
jdhoffa: Yes, but this will have to come in a future PR. 
The ALD->ABCD attack is going to be swift and focused. I'd rather not start changing it in some places, but not other, i think that will cause confusion.

R/target_market_share.R
```diff
@@ -282,105 +188,65 @@ target_market_share <- function(data,
...
 
   corporate_economy <- calculate_ald_benchmark(ald, region_isos, by_company)
 
-  relevant_sectors <- unique(data$sector)
-
-  relevant_corporate_economy <- corporate_economy %>%
-    filter(.data$sector %in% relevant_sectors)
+  relevant_corporate_economy <- filter(
+    corporate_economy,
+    .data$sector %in% unique(data$sector)
```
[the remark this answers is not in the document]
jdhoffa: Ya you're right, this is needlessly confusing.

R/target_market_share.R
jdhoffa: It should not have any `NA`s in it. In fact, there is a (very strict) check at the beginning of `target_market_share()` that causes the function to fail if `production` contains any `NA`s. We elected to go strict rather than flexible.

R/target_market_share.R
jdhoffa: It is a vector. I just default to using `data` as the first argument of most functions, but you are right, i will disambiguate.

R/target_market_share.R
jdhoffa: I was just refactoring. I removed the functions above, and bundled them in a different function that explains the purpose. Just trying to make the function body a little more readable/ clear.

R/target_market_share.R
jdhoffa: This is only used intermittently, and is removed later in a pivot_longer (or wider, I can't remember).

R/target_market_share.R
jdhoffa: I actually have no idea. @maurolepore added this `pivot_wider2` to solve a problem we ran into ages ago, and I truly don't remember what this problem was...

jdhoffa:
NITs addressed, merging now.

Issue #224, opened 2020-10-02
`target_sda` errors gracefully if `emission_factor` is wrong type


jdhoffa:
There has been an issue reported of incorrect formatting of some ald columns after reading them in with `readr`, causing the `target_sda` function to error. I wonder if this is because, in practice, the `ald` has many `NA`s in the `emission_factor` column, (potentially more than the first 1000 rows), meaning that `readr` may not know which data type to coerce the column-type to? 

In any case, we should potentially write a check that input data columns are the correct type, and pass a helpful error if not. @maurolepore what do you think? See the reprex below: 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(r2dii.analysis)

matched <- tibble::tribble(
  ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,            ~level, ~score,      ~sector,      ~name_ald, ~sector_ald,
  "L162",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive", "shaanxi auto",    "cement"
)

ald <- tibble::tribble(
  ~name_company,  ~sector, ~technology, ~year, ~production, ~emission_factor, ~plant_location, ~is_ultimate_owner,
  "shaanxi auto", "cement",    "cement",  2020,           1,                1,            "BF",               TRUE,
  "shaanxi auto", "cement",    "cement",  2021,           1,                2,            "BF",               TRUE,
  "shaanxi auto", "cement",    "cement",  2022,           1,                3,            "BF",               TRUE
  )

co2_intensity_scenario <- tibble::tribble(
  ~scenario,  ~sector,  ~region, ~year, ~emission_factor,           ~emission_factor_unit, ~scenario_source,
  "b2ds", "cement", "global",  2020,              0.6, "tons of CO2 per ton of cement",      "demo_2020",
  "b2ds", "cement", "global",  2050,              0.2, "tons of CO2 per ton of cement",      "demo_2020"
)

# bad format for emission_factor
bad_ald <- ald %>% 
  mutate(
    emission_factor = as.character(emission_factor)
  )

# works
out <- target_sda(
  matched,
  ald,
[...]

PR #388, opened 2022-03-28
WIP issue #224

julie-vienne:
Work in progress, review test-driven development with @jdhoffa

### DIFF
--- a/tests/testthat/test-target_sda.R
+++ b/tests/testthat/test-target_sda.R
@@ -616,3 +616,39 @@ test_that("doesn't output NAs if ald and scenario years are misaligned (#307,
     any(is.na(out$emission_factor_value))
   )
 })
+
+
+test_that("output useful error message when emission_factor is not of type double (#224)", {
+      matched <- tibble::tribble(
+       ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,            ~level, ~score,      ~sector,      ~name_ald, ~sector_ald,
+      "L162",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive", "shaanxi auto",    "cement"
+    )
+    bad_ald = tibble::tribble(
+      ~name_company,  ~sector, ~technology, ~year, ~production, ~emission_factor, ~plant_location, ~is_ultimate_owner,
+      "shaanxi auto", "cement",    "cement",  2020,           1,                1,            "BF",               TRUE,
+      "shaanxi auto", "cement",    "cement",  2021,           1,                2,            "BF",               TRUE,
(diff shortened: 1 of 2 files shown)


tests/testthat/test-target_sda.R
```diff
@@ -616,3 +616,39 @@ test_that("doesn't output NAs if ald and scenario years are misaligned (#307,
...
+    )
+    bad_ald <- ald %>%
+      mutate(
+        emission_factor = as.character(emission_factor)
+      )
+    out <- target_sda(
+      matched,
+      bad_ald,
+      co2_intensity_scenario
+    )
```
jdhoffa, line 645: You don't need to define the object `out` here, since you will call directly the function `target_sda()` within the function `expect_error`. Also note, in general, if you are expecting a function call to error, you should always be wrapping it in an `expect_error` call otherwise you will have tests failing that you expect to pass.

tests/testthat/test-target_sda.R
```diff
@@ -616,3 +616,39 @@ test_that("doesn't output NAs if ald and scenario years are misaligned (#307,
...
+    expect_error(
+      class = NULL,
```
jdhoffa, line 647: It is best to be specific about the class of the error message. You can do this with the function `rlang::abort()`, which is equivalent to `base::stop()`, but allows for classed conditions. 

What you can do is something along the lines of: 
``` r
if (condition_not_met){
abort("Condition isn't not met", class = "crucial_column_wrong_type") 
}
```

jdhoffa:
@julie-vienne so this is actually a very good WIP PR to demonstrate exactly why test-driven development (tdd) is so useful. 

My first comments are regarding the commit history of the PR. You have committed the changes to both the `R/target_sda.R` and `tests/testthat/test-target_sda.R` files in the same commit. I would rather suggest that you follow a commit workflow similar to: 
- commit 0: `feature-branch` is at the same commit as `main`
- commit 1: add a failing test
- commit 2: make necessary fixes so that failing test in commit 1 passes

This way, when I check out this PR locally, I can easily hop to commit 0, run tests, commit 1, run tests, and finally commit 2, and run tests. If I do that, I should see something like "all tests pass", "one test fails", "all tests pass". 

Right now, if you run the tests on this PR (either by `devtools::test()` or by the short-cut `ctrl+shift+T`), you will notice that there is a failure before the tests are able to run. 

In particular, I see the following output on the console: 
Screenshot 2022-03-29 at 12 41 53

To debug this, try opening a new PR, which has only the commit of the failing test in it for now, and then try running the tests again to see what output you get. (If you need some help with git, I can show you how to easily cherry-pick changes to one file into a new branch).

jdhoffa (changes requested):

cjyetman:
Might be good to point out that if/when you run `git merge main` it will likely drop you into a VIM editor, which is not exactly obvious or intuitive to someone that's never used VIM before.

PR #391, opened 2022-04-05
`target_sda` solves issue #224 with a helpful error message

julie-vienne:
Relates to PR #388 and @jdhoffa's comments

### DIFF
--- a/R/target_sda.R
+++ b/R/target_sda.R
@@ -72,6 +72,7 @@
 #'   )
 #'   out
 #' }
+
 target_sda <- function(data,
                        ald,
                        co2_intensity_scenario,
@@ -114,6 +115,7 @@ target_sda <- function(data,
   walk_(crucial_portfolio, ~ check_no_value_is_missing(data, .x))
 
   check_crucial_names(ald, crucial_ald)
+  check_type_emission_factor(ald)
 
   if (any(is.na(ald$emission_factor))) {
     warn(
@@ -199,6 +201,12 @@ target_sda <- function(data,
     adjusted_scenario,
     by_company = by_company
   )
+}
+
+check_type_emission_factor <- function(ald) {
+  if (!is.double(ald$emission_factor)) {
+  abort("The column emission_factor does not have the type double", class = "crucial_column_wrong_type")
+  }
 }
 
 check_unique_id <- function(data, column) {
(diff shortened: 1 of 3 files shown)


jdhoffa:
Nope. this is perfect. 
Checks and Tests are different. 

Almost all the github actions here run tests as one part of the process (except for Render README), and all of the Checks run tests. So this means if ay test fails, the whole check/ action will fail.

julie-vienne:
@jdhoffa all tests pass but the checks fail due to unstated dependencies about `tibble` (warning). I have however declared the imports  `tibble::tribble`. Any idea what's happening?

jdhoffa:
Yup, so because you are using `tibble` as it is re-exported from `dplyr`, you would need to namespace it like that (ie. `dplyr::tibble()`). 

But, on another note, check-out the utility functions `fake_matched()`, `fake_ald()` and `fake_co2_scenario()`. They are utility functions that we wrote and are really useful for iterating different versions of those datasets. We use them extensively in tests to quickly draft minimal datasets that expose different bugs, I think you should be able to re-write your test relatively easily with those functions, and in doing so, avoid using `tibble`. (Have a look at some of the other tests in the package to see how they are used :-D )

jdhoffa:
@julie-vienne just to explain the above comment, I triggered a github action by typing `/style`. These actions are defined here: https://github.com/2DegreesInvesting/r2dii.analysis/blob/main/.github/workflows/pr-commands.yaml

The `/style` command, basically just triggers a call to `styler::style_pkg()`, which checks that all code is styled (using the tidyverse style https://style.tidyverse.org/), and pushes a commit in case any of the files are not styled.

R/target_sda.R
```diff
@@ -201,6 +203,12 @@ target_sda <- function(data,
   )
 }
 
+check_type_emission_factor <- function(ald) {
+  if (!is.double(ald$emission_factor)) {
+  abort("The column emission_factor of the asset-level data frame (`ald`) does not have the type double", class = "crucial_column_wrong_type")
```
jdhoffa, line 208: NIT: No need to write out the full name of the dataset. Since it is an argument of the function, referring to it simply as `ald` is sufficient detail.

jdhoffa (approved):
One NIT comment, but otherwise LGTM :-)

PR #393, opened 2022-04-12
Plot percentage in units of "percent" (rather than a fraction)

### LINKED ISSUE #392: `percentage_of_initial_production_by_scope` outputs a fraction, and not a production
Notice in the reprex below, the difference between 125 and 120 is "-4%" and not "-0.04 percent". Thanks @MonikaFu 

``` r
library(tidyverse)
library(r2dii.data)
library(r2dii.analysis)

portfolio <- tibble::tribble(
  ~id_loan, ~loan_size_outstanding, ~loan_size_outstanding_currency, ~loan_size_credit_limit, ~loan_size_credit_limit_currency, ~id_2dii,            ~level, ~score,      ~sector, ~name_ald,  ~sector_ald,
  "L162",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive",   "comp1", "automotive",
  "L162",                      1,                           "EUR",                       2,                            "EUR",    "UP1", "ultimate_parent",      1, "automotive",   "comp2", "automotive"
)

ald <- tibble::tribble(
  "comp1", "automotive",  "electric",  2020,          10,                1,            "BF",               TRUE,
  "comp1", "automotive",       "ice",  2020,          30,                1,            "BF",               TRUE,
  "comp1", "automotive",  "electric",  2021,          20,                1,            "BF",               TRUE,
  "comp1", "automotive",       "ice",  2021,          20,                1,            "BF",               TRUE,
  "comp2", "automotive",  "electric",  2020,          90,                1,            "BF",               TRUE,
  "comp2", "automotive",       "ice",  2020,          95,                1,            "BF",               TRUE,
  "comp2", "automotive",  "electric",  2021,         100,                1,            "BF",               TRUE,
  "comp2", "automotive",       "ice",  2021,         100,                1,            "BF",               TRUE
)
scenario <- tibble::tribble(
  "sds", "automotive",  "electric", "global",  2020,     1,     0,      "demo_2020",
  "sds", "automotive",       "ice", "global",  2020,     1,     0,      "demo_2020",
  "sds", "automotive",  "electric", "global",  2021,  1.85,  0.34,      "demo_2020",
  "sds", "automotive",       "ice", "global",  2021,   0.6,  -0.2,      "demo_2020"
)

out <- target_market_share(
  portfolio,
  ald,
  scenario,
  region_isos_demo,
  by_company = FALSE,
  weight_production = FALSE
)

out %>% 
  filter(technology == "ice", metric == "projected") %>% 
  select(year, production, percentage_of_initial_production_by_scope)
#> # A tibble: 2 × 3
#>    year production percentage_of_initial_production_by_scope
#>   <dbl>      <dbl>                                     <dbl>
#> 1  2020        125                                      0   
#> 2  2021        120                                     -0.04
```

<sup>Created on 2022-04-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

jdhoffa:
Closes #392 Thanks @MonikaFu

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -237,8 +237,10 @@ add_percentage_of_initial_production_by_scope <- function(data,
     mutate(initial_sector_production = first(.data$sector_production)) %>%
     ungroup() %>%
     mutate(
-      percentage_of_initial_technology_production = (.data$production - .data$initial_technology_production) / .data$initial_technology_production,
-      percentage_of_initial_sector_production = (.data$production - .data$initial_technology_production) / .data$initial_sector_production
+      percentage_of_initial_technology_production =
+        (.data$production - .data$initial_technology_production) / .data$initial_technology_production * 100,
+      percentage_of_initial_sector_production =
+        (.data$production - .data$initial_technology_production) / .data$initial_sector_production * 100
     ) %>%
     mutate(
       scope = dplyr::case_when(
--- a/tests/testthat/test-target_market_share.R
+++ b/tests/testthat/test-target_market_share.R
@@ -1253,19 +1253,19 @@ test_that("w/ known input, outputs `percent_of_initial_production_by_scope` as
 
   expect_equal(
     out_percent$corporate_economy$percentage_of_initial_production_by_scope,
-    c(0.089, -0.040),
+    c(8.9, -4.0),
     tolerance = 1e-2
   )
 
   expect_equal(
     out_percent$projected$percentage_of_initial_production_by_scope,
-    c(0.089, -0.040),
+    c(8.9, -4.0),
     tolerance = 1e-2
   )
 
   expect_equal(
     out_percent$target_sds$percentage_of_initial_production_by_scope,
-    c(0.34, -0.40),
+    c(34, -40),
     tolerance = 1e-2
   )
 })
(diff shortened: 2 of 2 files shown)


cjyetman (approved):
My only really comment here is that -0.04 is equal to -4%. TBH, I prefer the data to be the real value that it is, and if one wants to format that as a percentage, that's easy enough to do at the last moment. Multiplying the values *100 to put them in percentage units seems like a bad idea to me, but I suppose in this case that's what truly matches the column name, so I guess it's appropriate here.

jdhoffa:
@cjyetman I tend to agree actually... this came from a conversation with @MonikaFu regarding what ends up getting plotted in `r2dii.plot` (and a desire to keep calculations, even very simple ones, out of that package). 

@MonikaFu what do you think?

MonikaFu:
So - either we make it a fraction but then change the name or we rutn the numbers into 'percentages'.

jdhoffa:
I feel like it's sort of ambiguous actually... like I personally wouldn't be surprised to see a column that is actually called `percent` and has units of percentage, but with values in the range (0,1)... It would almost surprise me more actually to see it have values in the range (0,100). Does that make sense?

jdhoffa:
> @jdhoffa - you are right. Hmmm... I don't know to be honest. There is not a right answer I think. We just need to make a decision. On my end, I kind of don't like the idea of a plotting function called `plot_*` doing something else than that. But if we all think that it is better to multiply by 100 there then I'll do it and probably give a warning to the user that the function does it.
> 
> Alternatively, I could use `scales::percent` but also not 100% sure if I like this idea that the value that you are plotting is always a percentage as in theory, you might sometimes want to plot the absolute 'production' values.

So my preference would definitely be to output the percentages on the range (0,1) and handle percentage formatting in the plotting... Maybe something we can bring up n the dev hangout? but we need to make a decision soonish

MonikaFu:
I will try to discuss this with @AlexAxthelm. Not opposed entirely to doing it in plotting but would like to consult with someone experienced before.

MonikaFu:
@jdhoffa - the decision has been made to keep the column between 0 and 1. So I guess that you can close this PR?

jdhoffa:
Closing!

PR #396, opened 2022-04-13
`target_sda` now outputs unweighted `emission_factor` by default, if `by_company = TRUE`

### LINKED ISSUE #376: Target_sda - Company level results wrong as they are loan weighted
In the target_sda function, there is no argument for Weight_production. (in the same way that there is for target_market_share) 

This means that when by_company is set to = TRUE the companies emission factors are weighted by their loan size. 
 
We want to be able to have the absolute (non-weighted) emission factors of companies when assessing at the company level. 

Hence the argument "weight_production" needs to be added to the function "target_sda"

furthermore, if Weight_production = TRUE - a warning message should be shown (i.e. the same one that is shown in the case of target_market_share)

jdhoffa:
Now:
* `target_sda(..., by_company = FALSE)` will calculate the loan-weighted emission factors, and aggregate them to portfolio level
* `target_sda(..., by_company = TRUE)` will output unweighted emission factors, and output them at company level

Note that the function `target_market_share()` has the argument `weight_production` which allows the user to manually control this behavior. I have elected not to include an analogous `weight_emission_factor` argument in this function, as the weighted but un-aggregated `emission_factor` value is not really interpretable or useful for the user, and would likely just cause confusion. 

Closes #376

### DIFF
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -180,7 +180,6 @@ summarize_weighted_emission_factor <- function(data, ..., use_credit_limit = FAL
 }
 
 summarize_unweighted_emission_factor <- function(data, ...) {
-
   data <- data %>%
     select(-c(
       .data$id_loan,
@@ -191,7 +190,6 @@ summarize_unweighted_emission_factor <- function(data, ...) {
     group_by(.data$sector_ald, .data$year, ...) %>%
     summarize(emission_factor_projected = mean(.data$emission_factor)) %>%
     ungroup()
-
 }
 
 calculate_weighted_loan_metric <- function(data, metric) {
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -179,6 +179,21 @@ summarize_weighted_emission_factor <- function(data, ..., use_credit_limit = FAL
     ungroup()
 }
 
+summarize_unweighted_emission_factor <- function(data, ...) {
+
+  data <- data %>%
+    select(-c(
+      .data$id_loan,
+      .data$loan_size_credit_limit,
+      .data$loan_size_outstanding
+    )) %>%
+    distinct() %>%
+    group_by(.data$sector_ald, .data$technology, .data$year, ...) %>%
+    summarize(emission_factor_projected = mean(.data$weighted_loan_emission_factor)) %>%
+    ungroup()
+
+}
+
 calculate_weighted_loan_metric <- function(data, metric) {
   crucial <- c(metric, "loan_weight")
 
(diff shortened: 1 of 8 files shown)


R/summarize_weighted_production.R
```diff
@@ -179,6 +179,19 @@ summarize_weighted_emission_factor <- function(data, ..., use_credit_limit = FAL
...
+summarize_unweighted_emission_factor <- function(data, ...) {
+  data <- data %>%
+    select(-c(
+      .data$id_loan,
+      .data$loan_size_credit_limit,
+      .data$loan_size_outstanding
+    )) %>%
+    distinct() %>%
+    group_by(.data$sector_ald, .data$year, ...) %>%
+    summarize(emission_factor_projected = mean(.data$emission_factor)) %>%
```
cjyetman, line 191: nit: you can use `.groups = "drop"` in `summarise()` and skip the `ungroup()`

tests/testthat/test-target_market_share.R
```diff
@@ -488,7 +488,7 @@ test_that("w/ known input, outputs `technology_share` as expected (#184, #262)",
     c(0.6875, 0.3125)
   )
 
-  FIXME <- 1e-2  # Do we need 1e-3?
+  FIXME <- 1e-2 # Do we need 1e-3?
```
cjyetman, line 491: nit: would rather see a useful function name and use FIXME in a comment

cjyetman (approved):
seems fine syntactically... left a few minor comments

R/summarize_weighted_production.R
jdhoffa: Thank you, I actually prefer to call `ungroup()` explicitly (for totally aesthetic reasons).
Personally, I find it much easier to look through a dplyr chain for matching `group_by()` and `ungroup()` calls, than to look for a `.groups = "drop"` nested in a function call.

tests/testthat/test-target_market_share.R
jdhoffa: Will address this also in a separate PR

R/summarize_weighted_production.R
cjyetman: fair enough, and agree it's easier to glance and pair up a `group_by()` with a `ungroup()`.... I kind of wish there were a `by` argument in `summarise()` so you could do something like...
```
data %>%
  summarise(emission_factor_projected = mean(.data$emission_factor), by = c(sector_ald, year))
```
instead of...
```
data %>%
  group_by(sector_ald, year, ...) %>%
  summarise(emission_factor_projected = mean(emission_factor)) %>%
  ungroup()
```

PR #397, opened 2022-04-14
`target_sda` gains `region_isos` argument

### LINKED ISSUE #323: target_sda needs a "region" input argument
target_sda is different to target_market_share in terms of its argument inputs in that target_sda does not take "region" as an input whereas target_marekt_share does.   

In the past when using the scenario input file for target_sda this has not been an issue as the scenario used (B2DS from the ETP_2017) only have targets at the global level. Hence there was no need for a regional input file. 
However, we now have new scenario input files to be used with the target_sda function that do have multiple regional benchmarks. in follows that a regional input argument is needed. 

NB - the r2dii.data::region_isos needs an update to accommodate for all these new scenarios  - this an issue in r2dii.data number 241

@jdhoffa @maurolepore @Antoine-Lalechere

jdhoffa:
`target_sda()` gains the argument `region_isos`. This will allow for regional granularity in the `SDA` output. 

One unfortunate note: 
* The function signature of `target_market_share` is: 
`target_market_share(data, ald, scenario, region_isos, use_credit_limit, by_company, weight_production)`

unfortunately, since we are adding `region_isos` to the `target_sda` function AFTER publishing the package, we cannot add the parameter in the same place as `target_market_share()`, it must be added as the final parameter (or else we risk breaking users old code). 

ie. the function of `target_sda()` is: 
`target_sda(data, ald, co2_intensity_scenario, use_credit_limit, by_company, region_isos)`
with `region_isos` being the final argument. 

Would love to hear @AlexAxthelm @cjyetman and @maurolepore opinions about this. Obviously it would be much nicer to keep some symmetry between these two functions. 

Closes #323

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,5 @@
 # r2dii.analysis (development version)
+* `target_sda()` gains `region` argument (#323).
 
 * `target_market_share()` now only outputs values for years that are in both 
   `ald` and `scenario` inputs (#394). 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,5 @@
 # r2dii.analysis (development version)
-* `target_sda()` gains `region` argument (#323).
+* `target_sda()` gains `region_isos` argument (#323).
 
 * `target_market_share()` now only outputs values for years that are in both 
   `ald` and `scenario` inputs (#394). 
--- a/R/target_sda.R
+++ b/R/target_sda.R
@@ -22,11 +22,9 @@
 #'   `weighted_production_value` at the portfolio-level. Set to `TRUE` to output
 #'   `weighted_production_value` at the company-level.
 #'
-#' @return  A tibble with the CO2 emissions factors attributed to
-#' the portfolio. These values include the portfolio's actual projected CO2
-#' emissions factors, the scenario pathway CO2 emissions factors and the SDA
-#' calculated portfolio target emissions factors (see column
-#' `emission_factor_metric`).
+#' @return A tibble including the summarized columns `emission_factor_metric` and
+#'   `emission_factor_value`. If `by_company = TRUE`, the output will also have
+#'   the column `name_ald`.
 #'
 #' @export
 #'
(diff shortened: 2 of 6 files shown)


NEWS.md
```diff
@@ -1,4 +1,5 @@
 # r2dii.analysis (development version)
+* `target_sda()` gains `region` argument (#323).
```
cjyetman, line 2: `region_isos` not `region`?

R/target_sda.R
```diff
@@ -131,20 +131,39 @@ target_sda <- function(data,
...
+  region_isos <- region_isos %>%
+    filter(source %in% unique(co2_intensity_scenario$scenario_source))
+
+  if (identical(nrow(region_isos), 0L)) {
+    warn("Found no matching regions for input scenario", class = "no_match")
+    return(empty_target_sda_output())
+  }
+
+  ald <- ald %>%
+    mutate(plant_location = tolower(.data$plant_location)) %>%
```
cjyetman, line 143: nit: Why are the country isos lower case? Technically, they are upper case, so if someone uses data with proper isos they don't match.

tests/testthat/_snaps/target_sda.md
```diff
@@ -3,38 +3,38 @@
...
-       2 cement  2021 projected                              2    
-       3 cement  2022 projected                              3    
-       4 cement  2020 corporate_economy                      1    
-       5 cement  2021 corporate_economy                      2    
-       6 cement  2022 corporate_economy                      3    
-       7 cement  2020 target_b2ds                            1    
-       8 cement  2021 target_b2ds                            0.978
-       9 cement  2022 target_b2ds                            0.956
-      10 cement  2023 target_b2ds                            0.933
+      # A tibble: 68 x 6
```
cjyetman, line 6: nit: are snapshots the most appropriate way of testing data frames? seems like formatting could easily cause false positive failures

cjyetman (approved):
syntactically, this seems fine to me (given the minor change to the note in NEWS)

NEWS.md
jdhoffa: good catch!

tests/testthat/_snaps/target_sda.md
jdhoffa: Not particularly, snapshots are a pretty brittle test, but they are left in as `target_sda` still has somewhat poor test coverage.

R/target_sda.R
jdhoffa: The contents of the data in `r2dii.data` is always formatted lowercase, so this line ensures that the `ald` input, which by default has `plant_location` uppercase, is formatted to lowercase, so that the join matches. 

I see the point though, if the user uses a different `region_isos` file, we would need to also force them lowercase. Hmm, I will make a new issue for this.

PR #399, opened 2022-04-21
Add `region_isos_demo` to `target_sda` calls in README

jdhoffa:
Since merging #397, all calls to `target_sda` in `README` and the function documentation lead to 0-row output. 

New argument `region_isos` must be set to `r2dii.data::region_isos_demo` to correctly output data.


cjyetman (approved):
nit: this PR is probably 90+% changes and deletions not related to the stated purpose, which makes it very hard to review

PR #400, opened 2022-04-21
`target_market_share` warns when input data has `NA`s

### LINKED ISSUE #304: Align how `target_market_share` and `target_sda` handle input `NA`s
Currently `target_market_share` will fail if a crucial column (e.g. `production`) has `NA` values,  while `target_sda` allows for `NA`s to be input, however it automatically filters them out. 

This decision for `target_sda` to handle `NA`s in this way was made since the input dataset provided by AR to the banks WILL automatically have `NA`s in the `emission_factor` column since certain sectors have no `emission_factors`. Rather than having the users filter a dataset provided by "us", we thought it would be good for the software to work "out of the box" with the data provided, but it might be worth revisiting this decision. 

As a user, it could be nice for both of these functions to handle `NA`s in the same way. 

See reprex below (and notice that `target_market_share` ERRORS, while `target_sda` WARNS): 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(r2dii.data)
library(r2dii.analysis)

matched <- tibble::tribble(
)

ald <- tibble::tribble(
  "shaanxi auto", "automotive",       "ice",  2025,           1,                1,            "BF",               TRUE,
  "shaanxi auto", "automotive",       "ice",  2025,           NA,                NA,            "BF",               TRUE
)


target_market_share(
  matched, 
  ald,
  scenario_demo_2020,
  region_isos_demo
)
#> Error: Column `production` must not contain any `NA`s.

target_sda(
  matched,
  ald,
  co2_intensity_scenario_demo,
)
#> Warning: Found no scenario data for input ald.
#> # A tibble: 0 x 4
#> # … with 4 variables: sector <chr>, year <int>, emission_factor_metric <chr>,
#> #   emission_factor_value <dbl>
```

<sup>Created on 2021-06-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

julie-vienne:
Currently `target_market_share` fails if a crucial column (e.g. `production` or `emission_factor`) has `NA` values, while `target_sda` allows for NAs to be inputted and filters them out.

This PR is to align both functions so that they handle `NA`s in the same, i.e. they both warn the user.

Closes #304


tests/testthat/test-target_market_share.R
```diff
@@ -170,6 +170,20 @@ test_that("w/ NAs in crucial columns, errors with informative message", {
...
+  ald <- fake_ald(production = c(1, NA))
+  scenario <- fake_scenario()
+
+  expect_warning(
+    target_market_share(
+      matched,
+      ald,
+      scenario,
+      region_isos_stable,
+      class = "na_input_data"))
```
jdhoffa, line 184: Test looks great, my only comment is that, for `target_sda`, the class of the warning is `na_emission_factor`. 
For symmetry, it would be nice if the class here was similar, ie, `na_production`.

julie-vienne: I had thought of that too but I was unsure of the class because from my understanding, both `production` and `emission_factor` could have `NA`s, right? This is why I chose "na_input_data", to include both. I do however understand the point about symmetry and I'm happy to make the change :)

jdhoffa: Ya, so we have some options here. 
The easiest would be to write two different warnings, one for `target_market_share` and one for `target_sda`, and give them the class `na_production` and `na_emission_factor`. But this is sorta WET code. 

The other option would be to find the common denominator between the behaviour we want in both function. For example, I think `na_input_data` is too general, since it isn't the behaviour we expect for NAs in ALL input datasets, it is only for two very special columns in the `ald` input dataset.

julie-vienne: Totally agree! Especially since I can then write a function, e.g. like `check_no_value_is_missing` in `utils.R` but with a warning and call that function in both `target_sda` and `target_market_share`

julie-vienne: something like `class = "na_input_ald"` or  `class = "na_ald_prod_ef"` ?

julie-vienne:
@jdhoffa so far `target_sda` errors when `production` has `NA`s. Have I understood correctly that it should now only warn and filter out the `NA`s, like for `emission_factor` in `target_sda` and `production` in `target_market_share`?

R/utils.R
```diff
@@ -9,6 +9,21 @@ check_no_value_is_missing <- function(data, column) {
...
+    warning_message = paste("Removing rows in", name_dataset, "where `{column}` is NA")
+    warn(
+      glue(warning_message),
+      class = "na_crucial_economic_input"
+    )
+
+    data <- filter(data, !is.na(.data[[column]]))
+
+    return(data)
+  }
```
jdhoffa, line 24: So the issue is that you need to put this `return(data)` outside of the `if()` statement. 

What is happening now is that, if the input `data` passes `anyNA`, the function works as expected, but if it doesn't, the function returns NULL, and that's why all the tests are failing.

julie-vienne: Makes sense, thanks for checking!

jdhoffa (approved):
lgtm, thanks julie! :-)

PR #401, opened 2022-04-22
Update to r-lib actions v2

### LINKED ISSUE #387: Upgrade to r-lib v2 Actions
Official v2 example actions here: https://github.com/r-lib/actions/tree/v2-branch/examples

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/check-full.yaml#L46-L58

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/check-full_r2dii-devel.yaml#L46-L64

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/pkgdown.yaml#L18-L24

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/pr-commands.yaml#L14-L53

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/render-readme.yaml#L19-L20

https://github.com/2DegreesInvesting/r2dii.analysis/blob/a54247b7db9f33e8e2bdd7662e7afce5bbdd1244/.github/workflows/test-coverage.yaml#L20-L24

jdhoffa:
See https://github.com/2DegreesInvesting/r2dii.data/pull/278 for more context.

Also added necessary dependencies to ensure pkgdown workflow passes.

Closes #387

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -1,33 +1,46 @@
-# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
     branches: [main, master]
-    tags: ['*']
+  pull_request:
+    branches: [main, master]
+  release:
+    types: [published]
+  workflow_dispatch:
 
 name: pkgdown
 
 jobs:
   pkgdown:
     runs-on: ubuntu-latest
+    # Only restrict concurrency for non-PR jobs
+    concurrency:
+      group: pkgdown-${{ github.event_name != 'pull_request' || github.run_id }}
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v2
 
-      - uses: r-lib/actions/setup-pandoc@v1
+      - uses: r-lib/actions/setup-pandoc@v2
 
-      - uses: r-lib/actions/setup-r@v1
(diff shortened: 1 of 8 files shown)


.github/workflows/pkgdown.yaml
```diff
@@ -1,33 +1,46 @@
...
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
         with:
           use-public-rspm: true
 
-      - uses: r-lib/actions/setup-r-dependencies@v1
+      - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: pkgdown
+          extra-packages: any::pkgdown, local::., any::tidyverse, any::r2dii.plot
```
cjyetman, line 33: nit: if you can specify only the specific pkgs needed instead of {tidyverse}, you can drastically improve the runtime of the action

cjyetman (approved):

.github/workflows/pkgdown.yaml
jdhoffa: Here again, we actually recommend to the user (in the website) to install and use the entire tidyverse. Not as a dependency, but just for ease of use of the rest of the r2dii packages. 
See here: https://2degreesinvesting.github.io/r2dii.analysis/articles/r2dii-analysis.html

PR #414, opened 2022-05-13
Update `target_*` functions to warn and change `isos` to lower-case in `region_isos`

julie-vienne:
In case the input `region_isos` has upper-cases in `isos`, the `target_market_share` and `target_sda` function will call a function to change the upper-cases to lower-cases and will issue a warning to the user. If this is not done, the joining of `region_isos` outputs 0 rows. 

Closes 398.

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -103,6 +103,8 @@ target_market_share <- function(data,
 
   abcd <- filter_and_warn_na(abcd, "production")
 
+  region_isos <- change_to_lowercase_and_warn(region_isos, "isos")
+
   warn_if_by_company_and_weight_production(by_company, weight_production)
 
   data <- ungroup(warn_grouped(data, "Ungrouping input data."))
--- a/R/target_sda.R
+++ b/R/target_sda.R
@@ -126,6 +126,8 @@ target_sda <- function(data,
   abcd <- filter_and_warn_na(abcd, "production")
   abcd <- filter_and_warn_na(abcd, "emission_factor")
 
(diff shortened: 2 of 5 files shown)


julie-vienne:
@jdhoffa so far, I've only created a test for target_market_share to make sure the function change_to_lowercase_and_warn updates and warns as it's supposed to. Do you think such a test is also necessary in target_sda as this function is also called there?

jdhoffa:
> @jdhoffa so far, I've only created a test for target_market_share to make sure the function change_to_lowercase_and_warn updates and warns as it's supposed to. Do you think such a test is also necessary in target_sda as this function is also called there?

Yes, I would say test both. Maybe this is a good time to elaborate a bit: 
The tests you write in the `testthat` folder are called "unit tests". A "unit test" tests one unit of the code, defined as "an exported function of the R package". So the goal should be: Test the final behavior of the exported function, in a way that is invariant to how it is actually implemented. (The reason this is desired, is so that you are free to refactor the code sometime down the road). 

In regards to the above, the function you wrote `change_to_lowercase_and_warn` is just one possible way of implementing the behavior, but if you ever wanted to refactor the code, and remove that function, you would want to do so in such a way that you're sure that the behavior is preserved in both `target_market_share` and `target_sda`. The only way to be sure of that is to have tested both of the functions explicitly

jdhoffa (changes requested):
LGTM, only change I would make is add the test also to `target_sda`

jdhoffa (approved):
lgtm

PR #417, opened 2022-09-19
Fix deprecated usage of `.data` in tidyselect functions

### LINKED ISSUE #416: Fixes for dev tidyselect
tidyselect no longer supports `.data`.
Relates to https://github.com/RMI-PACTA/r2dii.match/issues/417 and https://github.com/2DegreesInvesting/r2dii.analysis/pull/1

jdhoffa:
Relates to 2DegreesInvesting/r2dii.analysis#1
Closes #416

### DIFF
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -134,7 +134,7 @@ summarize_unweighted_production <- function(data, ..., with_targets = FALSE) {
         weighted_production_target = .data$production_target,
         .groups = "keep"
       ) %>%
-      ungroup(.data$technology) %>%
+      ungroup("technology") %>%
       mutate(
         weighted_technology_share = .data$weighted_production / sum(.data$weighted_production),
         weighted_technology_share_target = .data$weighted_production_target / sum(.data$weighted_production_target)
@@ -143,7 +143,7 @@ summarize_unweighted_production <- function(data, ..., with_targets = FALSE) {
   } else {
     data %>%
       summarize(weighted_production = .data$production, .groups = "keep") %>%
-      ungroup(.data$technology, .data$tmsr, .data$smsp) %>%
+      ungroup(all_of(c("technology", "tmsr","smsp"))) %>%
       mutate(weighted_technology_share = .data$weighted_production / sum(.data$weighted_production)) %>%
       group_by(!!!old_groups)
   }
--- a/R/summarize_weighted_production.R
+++ b/R/summarize_weighted_production.R
@@ -118,11 +118,9 @@ summarize_unweighted_production <- function(data, ..., with_targets = FALSE) {
   old_groups <- dplyr::groups(data)
 
   data <- data %>%
-    select(-c(
-      .data$id_loan,
-      .data$loan_size_credit_limit,
-      .data$loan_size_outstanding
-    )) %>%
+    select(
+      -all_of(c("id_loan", "loan_size_credit_limit", "loan_size_outstanding"))
+      ) %>%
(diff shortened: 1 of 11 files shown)


R/summarize_weighted_production.R
```diff
@@ -118,11 +118,9 @@ summarize_unweighted_production <- function(data, ..., with_targets = FALSE) {
...
 
   data <- data %>%
-    select(-c(
-      .data$id_loan,
-      .data$loan_size_credit_limit,
-      .data$loan_size_outstanding
-    )) %>%
+    select(
+      -all_of(c("id_loan", "loan_size_credit_limit", "loan_size_outstanding"))
+      ) %>%
```
cjyetman, line 123: nit: L123 is indented too far

R/summarize_weighted_production.R
```diff
@@ -145,7 +143,7 @@ summarize_unweighted_production <- function(data, ..., with_targets = FALSE) {
   } else {
     data %>%
       summarize(weighted_production = .data$production, .groups = "keep") %>%
-      ungroup(.data$technology, .data$tmsr, .data$smsp) %>%
+      ungroup(all_of(c("technology", "tmsr","smsp"))) %>%
```
cjyetman, line 146: pretty sure you should be able to use literal strings here without the `all_of()` and `c()` (`ungroup()` takes `...`), which seems less complicated, e.g.
``` r
tibble(a = 1, b = 2, c = 3) %>% 
  group_by(a, b) %>% 
  ungroup("a","b")
#> # A tibble: 1 × 3
#>       a     b     c
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
```

R/target_sda.R
```diff
@@ -298,11 +298,16 @@ compute_abcd_adjusted_scenario <- function(data, corporate_economy) {
...
+      all_of(
+        c(
+          "sector",
+          "scenario_source",
+          "region",
+          "emission_factor_corporate_economy"
+        )
+      )
     ) %>%
+    rename(baseline_emission_factor = "emission_factor_corporate_economy") %>%
```
cjyetman, line 310: apparently, you can still do renaming in `select()` without warpping in `all_of()`, which could eliminate L310, e.g. 
``` r
tibble(a = 1, b = 2, c = 3, d = 4) %>% 
  select(
    all_of(c("a", "b")),
    new_c = "c"
    )
#> # A tibble: 1 × 3
#>       a     b new_c
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
```

cjyetman (approved):
to be clear, I did not test this (R CMD check, etc.), but it looks good conceptually

comments are non-critical

R/target_sda.R
jdhoffa: yeah I noticed that too. I actually only separated them because I found it easier to read

R/summarize_weighted_production.R
jdhoffa: I agree that it's possible, I notice that you can actually do something similar with `select` (It doesn't complain if you pass it a normal vector anymore). 

In any case it's pretty unclear to me what the best practice approach is, so I'd rather just play it as safe as possible for now. 

(In this PR: https://github.com/2DegreesInvesting/r2dii.analysis/pull/1 , `ungroup()` calls `all_of` in some places)

jdhoffa: ^ yes totally, I noticed that as well, that's why it was still a little unclear to me

PR #437, opened 2023-06-05
update render readme yml

jdhoffa:


.github/workflows/render-readme.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v2
```
cjyetman, line 18:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/render-readme.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: macOS-latest
```
cjyetman, line 16: might be better (cheaper) to run this on ubuntu

.github/workflows/render-readme.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - uses: r-lib/actions/setup-r@v1
+      - uses: r-lib/actions/setup-r@v2
       - uses: r-lib/actions/setup-pandoc@v1
```
cjyetman, line 20:
```suggestion
      - uses: r-lib/actions/setup-pandoc@v2
```

.github/workflows/render-readme.yaml
jdhoffa: I believe it's all free since the package is public

jdhoffa:
Leaving as a draft for now, as I try to figure out why the yml isn't working.

jdhoffa:
Closing since stale

PR #438, opened 2023-07-05
Begin deprecation of `green_or_brown` in favour of `increasing_or_decreasing`

### LINKED ISSUE #434: Begin deprecation of usage of `green_or_brown`
The dataset `r2dii.data::green_or_brown` is being deprecated in favor of `increasing_or_decreasing` 
Relates to: https://github.com/RMI-PACTA/r2dii.data/issues/307 

From the perspective of `r2dii.analysis`, this affects the following functions: 
- `join_abcd_scenario`
- `summarize_weighted_production`
- `target_market_share`

This will be a bit of a tricky deprecation since the name, column name, and contents of the dataset are all changing. 

Deprecation will be done in two stages. Once `r2dii.data` is accepted on CRAN, with the new `increasing_or_decreasing` dataset, I will:

- [ ] Adapt the above functions to accept EITHER `green_or_brown` or `increasing_or_decreasing` (with a warning if the input dataset is `green_or_brown`), noting that this will involve temporarily allowing two different datasets with different value names
- [ ] If functions receive the `green_or_brown` dataset, convert the col_name to `increasing_or_decreasing`, and convert the `green` value -> `increasing` (`brown` value -> `decreasing`). Internally, I will aim to get the deprecation as far as possible.

This will be sufficient to get us to the next release. 

In the following release, we will complete the deprecation.

jdhoffa:
Closes #434

### DIFF
--- a/R/target_market_share.R
+++ b/R/target_market_share.R
@@ -19,9 +19,12 @@
 #' @param weight_production Logical vector of length 1. `TRUE` defaults to
 #' outputting production, weighted by relative loan-size. Set to `FALSE` to
 #' output the unweighted production values.
+#' @param increasing_or_decreasing A data frame like
+#' [r2dii.data::increasing_or_decreasing].
 #' @param ald `r lifecycle::badge('superseded')` `ald` has been superseded by
 #'   `abcd`.
-#' @param green_or_brown A data frame like [r2dii.data::green_or_brown]. 
+#' @param green_or_brown `r lifecycle::badge("superseded")` `green_or_brown` has
+#' been superseded by `increasing_or_decreasing`.
 #'
 #' @return A tibble including the summarized columns `metric`, `production`,
 #'   `technology_share`, `percentage_of_initial_production_by_scope` and
@@ -80,8 +83,9 @@ target_market_share <- function(data,
                                 use_credit_limit = FALSE,
                                 by_company = FALSE,
                                 weight_production = TRUE,
-                                green_or_brown = r2dii.data::green_or_brown,
-                                ald = deprecated()) {
+                                increasing_or_decreasing = r2dii.data::increasing_or_decreasing,
+                                ald = deprecated(),
+                                green_or_brown = deprecated()) {
   stopifnot(
     is.data.frame(data),
     is.data.frame(abcd),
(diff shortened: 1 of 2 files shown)


jdhoffa:
It is expected that some of these checks will fail until the latest version of https://github.com/RMI-PACTA/r2dii.data is on CRAN, since we are now depending on that latest dataset. As such, I will convert this to draft until that is the case. 

Also, note to self, this package will now depend on a latest version of `r2dii.data`

cjyetman:
Note in README please 😊

jdhoffa:
Wait a sec, I'm a dingus. 

The argument `green_or_brown` was only added in the development version of this package in the first place and never released, so don't need to deprecate it. Can just change it entirely to `increasing_or_decreasing`

jdhoffa:
Ok @cjyetman this is now ready for review. 

It was much easier than I anticipated, because `green_or_brown` was never actually an exported argument of any function in any of our CRAN releases, so I don't actually need to deprecate (I can just remove/ rename it as I wish)

cjyetman (dismissed):

cjyetman:
weird that so many checks failed?

jdhoffa:
> weird that so many checks failed?

This is because of all of the warnings due to the `many-to-many` relationship, which I handled in a separate PR.

jdhoffa:
> > weird that so many checks failed?
> 
> This is because of all of the warnings due to the `many-to-many` relationship, which I handled in a separate PR.

I have now merged that PR in. All checks should pass!

cjyetman:
looks like all "R-CMD-check r2dii-devel" tests pass, but all "R-CMD-check" tests fail. not sure what the difference is between the two

jdhoffa:
> looks like all "R-CMD-check r2dii-devel" tests pass, but all "R-CMD-check" tests fail. not sure what the difference is between the two

`devel` uses the development version of dependency packages (e.g. `r2dii.data` and `r2dii.match`) to run checks, whereas the normal one uses the CRAN release. 

This makes sense, since the functions now require the new dataset `increasing_or_decreasing`, which only exists in the development version of `r2dii.data` (it is not yet on CRAN). Once we release `r2dii.data` to CRAN, all checks should pass.

PR #440, opened 2023-07-19
Add test ensuring expected start year of `target_sda`

### LINKED ISSUE #439: SDA: first year of scenario should determine start of analysis
https://github.com/RMI-PACTA/r2dii.analysis/blob/cd3611cb90aa19b7f8e417b17ebe587223c8d121/R/target_sda.R#L366

For the `target_sda()` function the linked line shows that the scenario data is right joined onto the abcd. This means that in case the abcd has values starting in a year earlier than the first scenario year, this year will not be dropped from the analysis and the scenario thus does not determine the start year. Applying an inner join should fix this.

For comparison, in `target_market_share()` the following join is used and thus the first year of the scenario data determines the start year: https://github.com/RMI-PACTA/r2dii.analysis/blob/cd3611cb90aa19b7f8e417b17ebe587223c8d121/R/join_abcd_scenario.R#L65

**Reprex:**

```
library(tidyverse)
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

lbk_raw <- r2dii.data::loanbook_demo

# let us have the production data start a year befor the initial years in the scenario files
abcd <- r2dii.data::abcd_demo %>% 
  filter(year >= 2019)

matched <- r2dii.match::match_name(lbk_raw, abcd = abcd) %>% 
  r2dii.match::prioritize()

# min year for tmsr/smsp and sda sceanrios is the same
r2dii.data::scenario_demo_2020 %>% 
  pull(year) %>% 
  min()
#> [1] 2020

r2dii.data::co2_intensity_scenario_demo %>% 
  pull(year) %>% 
  min()
#> [1] 2020

# calculation for EI sectors, using SDA  
result_sda <- matched %>% 
  r2dii.analysis::target_sda(
    abcd = abcd,
    co2_intensity_scenario = r2dii.data::co2_intensity_scenario_demo,
    region_isos = r2dii.data::region_isos_demo
  )
#> Warning: Removing rows in abcd where `emission_factor` is NA

# calculation for production pathway sectors, using target market share
result_market_share <- matched %>% 
  r2dii.analysis::target_market_share(
    abcd = abcd,
    scenario = r2dii.data::scenario_demo_2020,
    region_isos = r2dii.data::region_isos_demo
  )

# comparison of intial years in results
result_sda %>% 
  pull(year) %>% 
  min()
#> [1] 2019

result_market_share %>% 
  pull(year) %>% 
  min()
#> [1] 2020
Created on 2023-07-19 with [reprex v2.0.2](https://reprex.tidyverse.org/)
```

jdhoffa:
Closes #439

### DIFF
--- a/tests/testthat/test-target_sda.R
+++ b/tests/testthat/test-target_sda.R
@@ -807,7 +807,7 @@ test_that("region_isos only has lowercase isos #398", {
   )
 })
 
-test_that("region_isos only has lowercase isos #398", {
+test_that("produces output with expected start years #439", {
 
   out <- target_sda(
       fake_matched(sector_abcd = "cement"),
--- a/tests/testthat/test-target_sda.R
+++ b/tests/testthat/test-target_sda.R
@@ -812,9 +812,11 @@ test_that("region_isos only has lowercase isos #398", {
   out <- target_sda(
       fake_matched(sector_abcd = "cement"),
       fake_abcd(sector = "cement", year = c(2024, 2025)),
-      fake_co2_scenario(year = 2024),
+      fake_co2_scenario(year = c(2025, 2050), emission_factor = c(0.5, 0.2)),
       region_isos = region_isos_stable
     )
 
-  expect_equal(min(out$year), 2025L)
+  out_relevant <- filter(out, grepl("target", emission_factor_metric))
+
+  expect_equal(min(out_relevant$year), 2025L)
 })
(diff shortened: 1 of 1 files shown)


tests/testthat/test-target_sda.R
```diff
@@ -806,3 +806,17 @@ test_that("region_isos only has lowercase isos #398", {
     )
   )
 })
+
+test_that("region_isos only has lowercase isos #398", {
```
cjyetman, line 810: seems like the test description is wrong, no?

jdhoffa: Ah yes, very much

jacobvjk (approved):

PR #441, opened 2023-07-19
Silence `many_to_many` warnings

### LINKED ISSUE #432: Quell warnings for many matches in new release of `dplyr`
Screenshot 2023-05-05 at 12 20 48 PM

jdhoffa:
Closes #432


cjyetman (approved):
I haven't yet used this option, but I believe this is how it works. I don't have context for what the data look like, but I assume you understand that the "many-to-many" relationship is not default for a reason, which implies one should carefully consider if that really should be happening and what the consequence of it is.

PR #443, opened 2023-08-25
Update GH actions

jdhoffa:


jdhoffa:
It is expected that only the `devel` checks pass until we release `r2dii.data` on CRAN.

PR #452, opened 2024-01-24
Add CODEOWNERS file

jdhoffa:
See https://github.com/RMI-PACTA/workflow.scenario.preparation/pull/2

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @jdhoffa will be requested for review when someone opens 
+# a pull request.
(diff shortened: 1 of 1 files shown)


jacobvjk:
@jdhoffa should we care about the failed check?

jdhoffa:
> @jdhoffa should we care about the failed check?

We should, but not here. This PR does not affect it. 

The render readme action seems to be broken. There is a conversation happening in parallel about it here:
https://github.com/RMI-PACTA/Actions/issues/11

I think in the future this kind of thing will fall in that repos domain, so not worrying about it here :-)

jacobvjk (commented):
lgtm

jacobvjk (approved):

PR #453, opened 2024-01-26
Add issues and PRs to maintainer project

jdhoffa:
See https://github.com/RMI-PACTA/r2dii.data/pull/321


cjyetman (approved):
The `github-token` is the hardest thing to get right here, which is not possible to verify through the git diff alone, but it looks like the action passed in the checks, so it looks good.

PR #455, opened 2024-01-26
Move r2dii.match from Suggests to Imports

### LINKED ISSUE #351: Rethink how we manage the dependency with r2dii.match
Relates to #352 

As of 2021-07-30 I see an error with debian-clang-devel at
https://cran.r-project.org/web/checks/check_results_r2dii.analysis.html

This particular error ultimately comes from an issue in stringdist, and the
maintainer seems to be working on it. However, this also makes me think
we are handling the dependency with r2dii.match in an odd way.

We new only suggest r2dii.match but we enforce it in examples. This is
contradictory. If we want to enforce r2dii.match then we should likley 
add it to Imports -- not Suggests. Or if we want users to opt out of
installing r2dii.match, then we should not throw an error if it's not 
installed. 

Given we expect users of r2dii.analysis to indeed use r2dii.match, I think
the simplest more reasanable soluiton is to add r2dii.match to Imports.

Alternatively we could keep r2dii.match in Suggests and either a) not
run the example where it's needed, or b) export a dataset from r2dii.analysis
that bypasses the need for r2dii.match.

jdhoffa:
Closes #351

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -49,13 +49,13 @@ Imports:
     lifecycle,
     magrittr,
     r2dii.data (>= 0.4.0),
+    r2dii.match,
     rlang (>= 0.1.2),
     tidyr,
     tidyselect,
     zoo
 Suggests: 
     covr,
-    r2dii.match,
     rmarkdown,
     roxygen2,
     spelling,


jdhoffa:
@jacobvjk could I get a review on this when you get a chance?

jacobvjk (approved):
lgtm

PR #462, opened 2024-01-31
Update roxygen

jdhoffa:


DESCRIPTION
```diff
@@ -65,4 +65,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.0
```
cjyetman, line 68: roxygen is already up to v7.3.1, if that matters to you

cjyetman (approved):

DESCRIPTION
jdhoffa: it does. Maybe i am out of date locally

cjyetman (approved):

PR #481, opened 2024-03-15
feat: `target_market_share` gains extended time-horizon for `target_*` outputs, to entire time-horizon of input `scenario`

### LINKED ISSUE #157: Allow for an extended timeline for `target_market_share()` targets
Unlink `target_sda`, the current output of `target_market_share()` is limited by the available `ald` data. 
For example, if `ald` data only exists until a certain year, say 2025, then all target output will only go until this year. 
This is due internally to how the data is joined together. 

As a bank, even if I am only able to see my portfolio's projected state until the max(year) of the ALD, I would like to see targets calculated for as long as there exists scenario projections. 

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(r2dii.data)
library(r2dii.match)
library(r2dii.analysis)

matched <- loanbook_demo %>%
  match_name(ald_demo) %>%
  prioritize()

# Calculate targets at portfolio level
out <- matched %>%
  target_market_share(
    ald = ald_demo,
    scenario = scenario_demo_2020,
    region_isos = region_isos_demo
  )
#> Joining, by = c("sector", "technology", "green_or_brown")

max(out$year)
#> [1] 2040

ald <- dplyr::filter(ald_demo, year == 2020)

out <- matched %>%
  target_market_share(
    ald = ald,
    scenario = scenario_demo_2020,
    region_isos = region_isos_demo
  )
#> Joining, by = c("sector", "technology", "green_or_brown")

max(out$year)
#> [1] 2020
```

<sup>Created on 2020-08-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jdhoffa:
Note: This required some significant changes to every exported function, e.g.:
`join_abcd_scenario`
`summarize_weighted_production`
`target_market_share`

A quick explanation of the changes that were made:
`join_abcd_scenario` - now outputs a dataset that:
* has a minimum `year` that is the shared minimum value of `year` between the `abcd` and `scenario` inputs
* has a maximum `year` that is the maximum `year` of `scenario` NOTE: this means that it fills in the `production` and `emission_factor` values as `NA`s, since it cannot infer them from the `abcd)

`summarize_weighted_production` - no longer complains if it receives an input with `NA` values in `production`
`target_market_share` had to have some things move around a bit, as the input data structure is now slightly expanded to include this new `NA` values of production. 

Closes #157


jdhoffa:
Going to mark this as draft again while I look into these failings checks

R/summarize_weighted_production.R
```diff
@@ -303,7 +302,7 @@ add_technology_share_target <- function(data) {
 
 check_zero_initial_production <- function(data) {
   companies_with_zero_initial_production <- data %>%
-    group_by(.data$technology, .data$name_abcd, .data$year) %>%
+    group_by(.data$technology, .data$name_abcd) %>%
```
jdhoffa, line 305: Apologies for this, it is a VERY random change in a bug I found in one of the checks for this function. 
I probably could have opened a separate PR for this, but it is truly just a change to a "check" (not even a change to the function logic itself), so I thought ok to just deal with it. 

The bug was identified when trying to render the `pkgdown` vignettes after opening this PR.

tests/testthat/test-join_abcd_scenario.R
```diff
@@ -259,3 +259,52 @@ test_that("outputs the same with upper/lower abcd$sector or abcd$technology", {
...
+      technology = c("hydrocap", "ice")
+      ),
+    region_isos = region_isos_stable
+  )
+
+  expect_equal(unique(out$sector_abcd), "power")
+
+})
+
+test_that("on fill in production timeline after t0 #157", {
```
jacobvjk, line 292: I don't understand what this expectation means. Looking at the test more closely it seems like the idea is to fill production only after an initial positive value, not before one. I am not sure I methodologically agree

R/summarize_weighted_production.R
```diff
@@ -187,10 +187,9 @@ summarize_unweighted_emission_factor <- function(data, ...) {
 }
 
 calculate_weighted_loan_metric <- function(data, metric) {
-  crucial <- c(metric, "loan_weight")
 
-  check_crucial_names(data, crucial)
-  walk_(crucial, ~ check_no_value_is_missing(data, .x))
+  check_crucial_names(data, c(metric, "loan_weight"))
+  walk_("loan_weight", ~ check_no_value_is_missing(data, .x))
```
jacobvjk, line 192: `calculate_weighted_loan_metric` is used on a number of metrics, not only `"production"`. Is it okay to remove `metric` in general from this expectation?

R/target_market_share.R
```diff
@@ -95,8 +93,23 @@ target_market_share <- function(data,
...
+  abcd <- dplyr::summarize(
+    abcd,
+    production = sum(.data[["production"]]),
+    .by = -"production"
+  )
+
   data <- aggregate_by_name_abcd(data)
 
+  if ("production" %in% colnames(scenario)) {
+    warn("The column `production` has been removed from the dataset `scenario`.
```
jacobvjk, line 106: how is this ever an issue?

jacobvjk (commented):
first set of comments. I will have to go through this again, especially the changes in `join_abcd_scenario()`

tests/testthat/test-join_abcd_scenario.R
jdhoffa: Firstly, noted this test is named horribly. I will go over all of the unit test names and try to make the more clear 😂 

To the methodological comment, the goal, as per George's request, is to "extend the timeline of the `target_*` outputs of the functions". 

Since target calculation only requires the production value at some t0 (either technology- or sector- total), this is doable for any arbitrary scenario timeline. 

We don't "fill in" the `production` at all.
Notice that, in the output of the function, for any years greater than what we have data for in the ABCD, the value of `production` is `NA`. 

So for example, the expectations for the output of `target_market_share` would be:
IF:
* `abcd` has production values for `2020:2025`
* `scenario` has target values for `2015:2030`
THEN:
* `out` for `metric = production` will have values between years `2020:2025`
* `out` for `metric = target_*` will have values between years `2020:2030`
* `out` will have no values prior to `2020` (the "start" year of production)

I think maybe this deserves a call to discuss, as I agree that the change is pretty drastic.

R/summarize_weighted_production.R
jdhoffa: Removing this expectation basically just means that and INPUT with `metric = NA` would give an output with `weighted_metric = NA`. Not necessarily ideal, but the other metrics calculated don't ever expect to have inputs with `NA`s. 

Is that OK, well I'm not sure. It shouldn't lead to any bugs if the input data is correct.
It may lead to a more difficult user experience if the input data is incorrect (e.g. one of the error messages will no longer be present).

R/target_market_share.R
jdhoffa: Not sure. I didn't add this check in this PR, i just refactored it from here: 
https://github.com/RMI-PACTA/r2dii.analysis/blob/170d4b0d60a3514607a962d32d2a848baaf298e9/R/target_market_share.R#L112-L118

It may be superfluous, but I guess when it was introduced there was some reason for it.

jdhoffa: Was introduced in this PR: https://github.com/RMI-PACTA/r2dii.analysis/pull/406

jdhoffa: Happy to explore if it's necessary, but I'd rather do so in a separate issue/ PR.

tests/testthat/test-join_abcd_scenario.R
jacobvjk: https://github.com/RMI-PACTA/r2dii.analysis/issues/482

R/summarize_weighted_production.R
jacobvjk: I guess one option would be to provide a boolean `check_metric` that allows toggling that check for the metric in each call. in that case we would not lose the check for all other metrics beside `"production"`, which seems the only one that should not use the check? also not a nicer interface, but it seems like a less drastic change in checks overall

jacobvjk (approved):
NB: bit unsure if it's better to remove the `check_crucial_names` for more variables than `"production"` or of it would be preferable to just add a switch for that. For users it won't make a difference. But I can see the point in not making the code base more complicated than necessary too.

R/summarize_weighted_production.R
jdhoffa: Done!

jacobvjk (approved):
lgtm

PR #532, opened 2025-02-14
update data dictionary to include definitions

cjyetman:
- depends on #529 
- depends on #533

⚠️ @jacobvjk should merge this when ready to ensure we don't conflict with the CRAN release versioning


jacobvjk (approved):
lgtm

PR #534, opened 2025-02-14
Update tests of deprecated functions

jacobvjk:
* updates tests of deprecated functions following https://lifecycle.r-lib.org/articles/communicate.html#deprecate-a-function (but using expect_warning instead of adding a snapshot test)
* turns the deprecated function `join_abcd_scenario()` into a wrapper around the new internal function `join_abcd_scenario_()` (following the same structure applied for `summarize_weighted_production()`). This allows quelling deprecation warnings in the tests of the exported function while avoiding warnings being thrown when the internal function is called from within `target_market_share()`

### DIFF
--- a/R/join_abcd_scenario.R
+++ b/R/join_abcd_scenario.R
@@ -35,22 +35,35 @@
 #' library(r2dii.match)
 #'
 #' valid_matches <- match_name(loanbook_demo, abcd_demo) %>%
-#' # WARNING: Remember to validate matches (see `?prioritize`)
+#'   # WARNING: Remember to validate matches (see `?prioritize`)
 #'   prioritize()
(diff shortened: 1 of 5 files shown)


cjyetman:
@jacobvjk fyi... the failing codecov test will be fixed by #535

cjyetman (approved):
pretty hard to follow and/or know if you managed to catch everything (not your fault, just the nature of it), but it looks good to me and the tests pass

jacobvjk:
> pretty hard to follow and/or know if you managed to catch everything (not your fault, just the nature of it), but it looks good to me and the tests pass

yea I know, it's unfortunate there are so many instances that are affected

